PR #1, opened 2017-08-28
Check error before defer commands that may fail.

pallavagarwal07:
If a file is opened, it returns the file pointer and error. However,
in case of an error, the file pointer would be nil. Thus, the error
needs to be checked, or else `defer file.Close()` may be executed on
a nil pointer causing the program to crash.
Same goes for other function calls like fuse.Mount().

This fixes the bug where sandboxfs would crash on a non-existent mountpoint.

### DIFF
--- a/integration/reconfiguration_test.go
+++ b/integration/reconfiguration_test.go
@@ -170,32 +170,41 @@ func TestReconfiguration_StreamFileDoesNotExist(t *testing.T) {
 	}
 	defer os.RemoveAll(tempDir)
 
-	mountPoint, err := ioutil.TempDir("", "mountPoint")
-	if err != nil {
-		t.Fatalf("failed to create temporary directory: %v", err)
-	}
-	defer os.RemoveAll(mountPoint)
+	nonExistentFile := filepath.Join(tempDir, "non-existent/fifo")
 
-	inFifo := filepath.Join(tempDir, "input")
-	inFile, err := os.OpenFile(inFifo, os.O_CREATE, 0666)
-	if err != nil {
-		t.Fatalf("Creating %q failed with error: %v", inFifo, err)
+	tests := []struct {
+		flag       string
+		wantStderr string
+	}{
+		{
+			"--input=" + nonExistentFile,
+			fmt.Sprintf("Unable to open file \"%s\" for reading: "+
+				"open %s: no such file or directory",
+				nonExistentFile, nonExistentFile),
+		},
+		{
+			"--output=" + nonExistentFile,
+			fmt.Sprintf("Unable to open file \"%s\" for writing: "+
+				"open %s: no such file or directory",
+				nonExistentFile, nonExistentFile),
+		},
 	}
-	inFile.Close()
 
-	outFifo := filepath.Join(tempDir, "non-existent/output")
-	wantStderr := fmt.Sprintf("Unable to open file \"%s\" for writing: open %s: no such file or directory",
-		outFifo, outFifo)
+	mnt := filepath.Join(tempDir, "mountPoint")
+	mkdirAllOrFatal(t, mnt, 0755)
+	defer os.RemoveAll(mnt)
 
-	stdout, stderr, err := runAndWait(1, "dynamic", "--input="+inFifo, "--output="+outFifo, mountPoint)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if len(stdout) > 0 {
-		t.Errorf("got %s; want stdout to be empty", stdout)
-	}
-	if !matchesRegexp(wantStderr, stderr) {
-		t.Errorf("got %s; want stderr to match %s", stderr, wantStderr)
+	for _, test := range tests {
+		stdout, stderr, err := runAndWait(1, "dynamic", test.flag, mnt)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if len(stdout) > 0 {
+			t.Errorf("got %s; want stdout to be empty", stdout)
+		}
+		if !matchesRegexp(test.wantStderr, stderr) {
+			t.Errorf("got %s; want stderr to match %s", stderr, test.wantStderr)
+		}
 	}
 }
 
--- a/integration/reconfiguration_test.go
+++ b/integration/reconfiguration_test.go
@@ -163,6 +163,42 @@ func TestReconfiguration_ExplicitStreams(t *testing.T) {
 	doReconfigurationTest(t, state, input, output)
 }
 
+func TestReconfiguration_StreamFileDoesNotExist(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "test")
+	if err != nil {
+		t.Fatalf("failed to create temporary directory: %v", err)
+	}
+	defer os.RemoveAll(tempDir)
+
+	mountPoint, err := ioutil.TempDir("", "mountPoint")
+	if err != nil {
+		t.Fatalf("failed to create temporary directory: %v", err)
+	}
+	defer os.RemoveAll(mountPoint)
+
+	inFifo := filepath.Join(tempDir, "input")
+	inFile, err := os.OpenFile(inFifo, os.O_CREATE, 0666)
+	if err != nil {
+		t.Fatalf("Creating %q failed with error: %v", inFifo, err)
+	}
+	inFile.Close()
+
+	outFifo := filepath.Join(tempDir, "non-existent/output")
+	wantStderr := fmt.Sprintf("Unable to open file \"%s\" for writing: open %s: no such file or directory",
+		outFifo, outFifo)
+
+	stdout, stderr, err := runAndWait(1, "dynamic", "--input="+inFifo, "--output="+outFifo, mountPoint)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(stdout) > 0 {
+		t.Errorf("got %s; want stdout to be empty", stdout)
+	}
+	if !matchesRegexp(wantStderr, stderr) {
+		t.Errorf("got %s; want stderr to match %s", stderr, wantStderr)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


jmmv:
Thanks!

This change is indeed fixing the bug where sandboxfs crashes on a non-existent mount point... but it's also fixing the handling of the `--input` and `--output` flags and no tests are affected by the latter. Which means our tests are incomplete. Therefore, either take those fixes out of this PR as they'd be unrelated, or extend the tests accordingly.

pallavagarwal07:
I'll add more tests then :)

pallavagarwal07:
I checked, but this is not possible to check in tests. The `--input` `--output` flags won't crash because the .Close() functions return `error` instead of panicking.
Even though it means that the corresponding change is unnecessary, I'd like to keep it for consistency.

jmmv:
Interesting that you can invoke an operation on a nil pointer and the operation doesn't crash. A bit of research indicates that this is intentional (but... yikes).

That does not mean this is not testable. The current code has no tests for invalid conditions in general (and `--input=/file/in/non-existent/directory` in particular) and having these is equally important to having tests for "good conditions". What you are observing could be undefined behavior but there is no way of knowing because the tests are not looking for it.

We need to have *many* more tests that validate invalid input conditions; whether they can or cannot trigger the current bug is separate (and I don't have ideas on how make the bug observable, but it doesn't matter much).

integration/reconfiguration_test.go
```diff
@@ -163,6 +163,42 @@ func TestReconfiguration_ExplicitStreams(t *testing.T) {
 	doReconfigurationTest(t, state, input, output)
 }
 
+func TestReconfiguration_StreamFileDoesNotExist(t *testing.T) {
```
jmmv: Which stream file does not exist? You are conflating two separate code paths here, which means one is not being tested. There are two different flags, so we need two have two disjoint tests. (A table-driven test can be sufficient here, where the test parameters are the flag name and the expected error message.)

integration/reconfiguration_test.go
```diff
@@ -163,6 +163,42 @@ func TestReconfiguration_ExplicitStreams(t *testing.T) {
...
+	inFifo := filepath.Join(tempDir, "input")
+	inFile, err := os.OpenFile(inFifo, os.O_CREATE, 0666)
+	if err != nil {
+		t.Fatalf("Creating %q failed with error: %v", inFifo, err)
+	}
+	inFile.Close()
+
+	outFifo := filepath.Join(tempDir, "non-existent/output")
+	wantStderr := fmt.Sprintf("Unable to open file \"%s\" for writing: open %s: no such file or directory",
+		outFifo, outFifo)
```
jmmv: nit: Why the line break? This is not done anywhere except for comments.

integration/reconfiguration_test.go
```diff
@@ -163,6 +163,42 @@ func TestReconfiguration_ExplicitStreams(t *testing.T) {
...
 }
 
+func TestReconfiguration_StreamFileDoesNotExist(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "test")
+	if err != nil {
+		t.Fatalf("failed to create temporary directory: %v", err)
+	}
+	defer os.RemoveAll(tempDir)
+
+	mountPoint, err := ioutil.TempDir("", "mountPoint")
```
jmmv: Why is mountPoint created with TempDir and not just using mkdirAllOrFatal under tempDir?

Does it even matter that this directory exists? I think sandboxfs will fail earlier due to the bad flags, so there is no need for this.

Also, everywhere else the directory is simply called "mnt".

pallavagarwal07: I was waiting for your opinion on this. While I'm still hazy on the test suite details, I still don't believe that asserting the assumption 
> Does it even matter that this directory exists? I think sandboxfs will fail earlier due to the bad
> flags, so there is no need for this.

is correct. Tests should not depend on the order in which our program checks the correctness of mountpoint/flags. In my opinion, the tests should check for a failure path when all else is correct (I have noticed similar inconsistencies in other present tests too).
If flag is incorrect **and** mountpoint is incorrect, why should a test decide that only one of them should show up as error and not the other one?

jmmv: That's a good point. But here is how I see this:

The fact that flags are validated first and that the mount point is used later is *user-observable behavior*. It can even be rationalized by saying that it makes no sense to try to mount the file system until all settings that affect it have been validated (which is how I think about it too).

Users themselves could rely on this undocumented behavior -- and if sandboxfs' behavior changed, their code would break. Maybe it makes our tests more fragile, but if our tests break eagerly as well, it's good because we can preemptively notice and address such behavioral change.

Note also that creating the mount point is essentially the same as adding unreachable code. There is no guarantee that what you wrote is correct because it has no effect on the behavior test. You'd have a typo in the directory you created and the test would still pass (too trivial to happen, right? Well... just wait for the code to evolve over many changes by different people and then it's not so obvious).

pallavagarwal07: > It can even be rationalized by saying that it makes no sense to try to mount the file system until all settings that affect it have been validated (which is how I think about it too).

The order could very well be:
1. Check the format of all flags.
2. Check existence of mount point and permissions (raise error?).
3. Open streams given by user in flags (raise error?).
4. Actually mount.

Which seems as natural to me as what happens currently.

> Users themselves could rely on this undocumented behavior -- and if sandboxfs' behavior changed, their code would break. 

Leaving aside the fact that I cannot think of any possible scenario where a user might depend on this behavior, I think that a user shouldn't/mustn't depend on this behavior. Writing a test that depends on this behavior is only encouraging a user to do so too (same about the recently pushed platform dependent test -- bug).

> There is no guarantee that what you wrote is correct because it has no effect on the behavior test. You'd have a typo in the directory you created and the test would still pass..

So to avoid the condition where we have a wrong situation in case we make a mistake (typo), we want to have a wrong situation as default?

I *really* think that a test checking for error condition should invoke exactly one error. It's very un-intuitive to me otherwise.

philwo: It's late, I'm at home already, maybe I'm missing something here. But the test looks totally fine to me?

To be honest, I don't even understand what you two are arguing about. :)

The test is creating a tempDir (which it needs as a space to put the inFifo / outFifo paths).
It creates a mountPoint, because, well, you need one when you call sandboxfs.

The fact that something ("some part of flag validation") happens before something else ("checking that the mountPath exists") and that's why we could in theory also use a non-existing mountPoint, because hey, YAGNI, feels like a way-too-clever hack to me rather than a well thought through concept - also because no one will even *get* what this is trying to test and on what internal behaviour this relies. Just make it simple and think of the maintainers in the future who might not understand why this test passes an invalid mountPath, when it actually wants to test what happens when you pass invalid flags.

Just have two tests, one for flag validation (with a valid mountPoint) and one for invalid mountPoints (but with valid flags)?

jmmv: I agree with your general premise, but I don't think it applies here and making it apply just makes the tests more complex for no reason.

> The order could very well be:
> 
> 1. Check the format of all flags.
> 1. Check existence of mount point and permissions (raise error?).
> 1. Open streams given by user in flags (raise error?).
> 1. Actually mount.
>
> Which seems as natural to me as what happens currently.

This may "sound natural" but, other than involving extra unnecessary work in the code (step 2), this approach is buggy! https://blogs.msdn.microsoft.com/jaredpar/2009/12/10/the-file-system-is-unpredictable/ . So no, this order doesn't make sense.

You really can only use the mount point after everything else is valid, because you can only use the mount point via the actual mount operation. This implies that not creating the mount point is fine and doesn't make this test wrong. You can't reach that condition, which is kinda how it should be.

> Leaving aside the fact that I cannot think of any possible scenario where a user might depend on this behavior, I think that a user shouldn't/mustn't depend on this behavior.

This is less important than what I mentioned above, but still: it doesn't matter what you think about what users will do: https://twitter.com/onoffleftright/status/885627206033997825 (sorry for the bad reference; you may be able to find the talk). User-visible behavior will be relied upon by others, so it ought to be tested for regressions -- especially in this case where the code is doing the only reasonable thing it can do.

jmmv: There is one last consideration I wanted to raise on this topic. It is good practice for any piece of code to avoid modifying outputs until all inputs have been dealt with, and inputs should only be touched after all of them have been validated. In other words: first, validate; second, read; third, modify. This order should be enforced so that side-effects are minimized to happen only when everything is known to be in a good state.

In the case of a running a file system, the obvious "output" is the mount point. Everything else is an input. An easy way to enforce the above properties is simply to *not* create the mount point during error testing. This ensures that the error you get back corresponds to the input validation, and that the output was not touched at all during that process.

integration/reconfiguration_test.go
```diff
@@ -163,6 +163,51 @@ func TestReconfiguration_ExplicitStreams(t *testing.T) {
...
+				"open %s: no such file or directory",
+				nonExistentFile, nonExistentFile),
+		},
+	}
+
+	mnt := filepath.Join(tempDir, "mountPoint")
+	mkdirAllOrFatal(t, mnt, 0755)
+	defer os.RemoveAll(mnt)
+
+	for _, test := range tests {
```
jmmv: Please use `t.Run()` to run subtests so that they are independent.

Also please be consistent with the naming and style used everywhere else. `cli_test.go` and `signal_test.go` contain a bunch of samples.

pallavagarwal07: Okay..
In the meantime could you check why Travis is still failing on OSX? I don't have a Mac to test the code.

jmmv: I can do the proposed changes if you want prior internal review+checkin, and test them against macOS manually. Let me know.

pallavagarwal07: I think I have already done the changes. But please do test them on macOS manually. I have no way to do so, and Travis build queues are too long to be of much use.

PR #3, opened 2017-09-09
Migrate code to use /x/sys/unix instead of syscall wherever possible.

pallavagarwal07:
According to https://golang.org/pkg/syscall/, the syscall package is now locked down and existing code should be ported to use /x/sys/unix whenever possible.
References to Stat_t still use the old syscall package due to values returned by existing library packages.

### DIFF
--- a/integration/common.go
+++ b/integration/common.go
@@ -27,11 +27,11 @@ import (
 	"reflect"
 	"regexp"
 	"strings"
-	"syscall"
 	"testing"
 	"time"
 
 	"bazil.org/fuse"
+	"golang.org/x/sys/unix"
 )
 
 const (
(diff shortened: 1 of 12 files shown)


jmmv:
From a cursory look, it seems that you could get rid of the `syscall.Stat_t` references if you replaced `os.Lstat` with `unix.Lstat`?

(This would still leave behind the `syscall.*` references for signals, but that's still a known deficiency of the `unix` package.)

jmmv:
I think it'd be nice to migrate to the `unix` package... and to do so everywhere possible. Remaining usages of the `syscall` package should be annotated with a TODO pointing at an upstream bug report that tracks the implementation of the missing features (e.g. signal handling).

However... I'm going to close this PR as stale given that the codebase has gained several more usages of `syscall` that would need addressing and this cannot be merged. Apologies for the delay.

PR #6, opened 2017-11-18
Add Bazel build scripts, source code linting, and IDE integration

jmmv:
This set of commits adds a Bazel build for sandboxfs and then uses that to add source code linting (with pre-commit checks) and IDE integration with VSCode.

These commits are "clean" on their own, so they shall be merged without squashing.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -23,8 +23,23 @@ To get started:
         and remove the `go_version = "host"` parameter from the
         `go_register_toolchains` statement.
 
-1.  Run `bazel build //cmd/sandboxfs` to download all required dependencies
-    and build the final binary.
+1.  Run `bazel build //admin/install` to download all required dependencies,
+    build the final binary, and build the installation tool.
+
+1.  To copy all binary and data files to the default destination of
+    `/usr/local`, run `./bazel-bin/admin/install/install`.
+
+    *   You will (most likely) need superuser permissions to install
+        under `/usr/local`, so run the previous command with `sudo`.
+
+    *   The reason the above command is not `sudo bazel run ...` is because
+        running Bazel under `sudo` will cause a full rebuild of everything.
+        Don't run Bazel as root.  Instead, just build the installation tool
+        first and run it separately, as described above.
+
+    *   If you want to install sandboxfs under a custom prefix, run
+        `./bazel-bin/admin/install/install --prefix=/path/to/prefix`
+        instead.
 
 ## From sources with the Go tools
 
@@ -33,3 +48,7 @@ standard Go tools as follows:
 
     go get github.com/bazelbuild/sandboxfs/cmd/sandboxfs
     go build github.com/bazelbuild/sandboxfs/cmd/sandboxfs
+
+This is not recommended because this mechanism does not provide a way to
+install the built product.  (Yes, `go install` does install the binary but does
+not handle other support files such as documentation.)
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,35 @@
+# Installation instructions
+
+Given that there have not yet been any formal releases of sandboxfs, there
+currently are no prebuilt binaries available.  To use sandboxfs, you will
+have to build and install it from a fresh checkout of the GitHub tree.
+
+## From sources with Bazel
+
+The preferred mechanism to build sandboxfs is to use the
+[Bazel](http://bazel.build) build system.  The reason is that this provides
+a reproducible build environment with pinned versions of all required
+dependencies and also integrates well with various support tools used
+during installation and during development.
+
+To get started:
+
+1.  [Download and install Bazel](https://bazel.build/).
+
+1.  [Download and install Go](https://golang.org/).
+
+    *   If you don't want to do this by hand and would rather have Bazel
+        manage the installation of Go for you, edit the `WORKSPACE` file
+        and remove the `go_version = "host"` parameter from the
+        `go_register_toolchains` statement.
+
+1.  Run `bazel build //cmd/sandboxfs` to download all required dependencies
+    and build the final binary.
+
+## From sources with the Go tools
+
+You can also build and install sandboxfs from this Git repository using the
+standard Go tools as follows:
+
+    go get github.com/bazelbuild/sandboxfs/cmd/sandboxfs
+    go build github.com/bazelbuild/sandboxfs/cmd/sandboxfs
--- a/admin/install/install.go
+++ b/admin/install/install.go
@@ -1,0 +1,120 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License.  You may obtain a copy
+// of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+// License for the specific language governing permissions and limitations
+// under the License.
+
+// The install binary installs sandboxfs and all of its support files.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/bazelbuild/sandboxfs/internal/shell"
+)
+
+const (
+	// Relative path to the runfiles directory for the install script.  Used if the script is
+	// run by hand from the workspace directory, without using "bazel run".
+	runfiles = "bazel-bin/admin/install/install.runfiles/sandboxfs"
+
+	// Path to the built sandboxfs binary relative to the runfiles directory.
+	sandboxfsBin = "cmd/sandboxfs/sandboxfs"
+
+	// Desired permissions for installed files.
+	dataPerm = 0644
+	execPerm = 0755
+	dirPerm  = 0755
+)
+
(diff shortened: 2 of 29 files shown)


INSTALL.md
```diff
@@ -0,0 +1,43 @@
...
+1.  [Download and install Bazel](https://bazel.build/).
+
+1.  [Download and install Go](https://golang.org/).
+
+    *   If you don't want to do this by hand and would rather have Bazel
+        manage the installation of Go for you, edit the `WORKSPACE` file
+        and remove the `go_version = "host"` parameter from the
+        `go_register_toolchains` statement.
+
+1.  Run `bazel run //admin/install` to download all required dependencies,
```
rupertks: The install failed for me (on a Mac). Do we expect users to specify a custom prefix, or to have to use sudo, or ?

```
$ bazel run //admin/install
INFO: Found 1 target...
INFO: From SkylarkAction external/go_stdlib_darwin_amd64_cgo/bin/go [for host]:
# runtime/cgo
clang: warning: argument unused during compilation: '-pthread' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-gno-record-gcc-switches' [-Wunused-command-line-argument]
...
INFO: Elapsed time: 43.372s, Critical Path: 25.90s

INFO: Running command line: bazel-bin/admin/install/install
2017/11/20 13:08:24 cmd/sandboxfs/sandboxfs -> /usr/local/bin/sandboxfs (mode -rwxr-xr-x)
ERROR: failed to copy cmd/sandboxfs/sandboxfs to /usr/local/bin/sandboxfs: cannot open /usr/local/bin/sandboxfs for write: open /usr/local/bin/sandboxfs: permission denied
ERROR: Non-zero return code '1' from command: Process exited with status 1.
````

Comment (jmmv) [reply INSTALL.md]: Oh right... yes, you need root to copy onto `/usr/local` obviously, but `sudo bazel` is a pretty bad idea. I've updated the instructions accordingly.

Thanks for trying this instead of just reviewing the code! :-)

Comment (rupertks) [reply INSTALL.md]: Thanks!

I just tried following the new INSTALL.md, here's the new error I got:

```
$ bazel build //admin/install
INFO: Found 1 target...
Target //admin/install:install up-to-date:
  bazel-bin/admin/install/install
INFO: Elapsed time: 12.385s, Critical Path: 2.08s

$ ./bazel-bin/admin/install/install
2017/11/22 14:34:18 cmd/sandboxfs/sandboxfs -> /usr/local/bin/sandboxfs (mode -rwxr-xr-x)
ERROR: failed to copy cmd/sandboxfs/sandboxfs to /usr/local/bin/sandboxfs: cannot open cmd/sandboxfs/sandboxfs for read: open cmd/sandboxfs/sandboxfs: no such file or directory

$ sudo ./bazel-bin/admin/install/install
Password:
2017/11/22 14:34:24 cmd/sandboxfs/sandboxfs -> /usr/local/bin/sandboxfs (mode -rwxr-xr-x)
ERROR: failed to copy cmd/sandboxfs/sandboxfs to /usr/local/bin/sandboxfs: cannot open cmd/sandboxfs/sandboxfs for read: open cmd/sandboxfs/sandboxfs: no such file or directory
```

Comment (jmmv) [reply INSTALL.md]: OK... let's try again.

I've pushed a fix -- as a new commit to avoid the problem you observed before in GitHub and not being able to see what changed. However, this new commit will need fixing up into the right commit pre-merge.

Comment (rupertks) [reply INSTALL.md]: Okay, that worked!

```
$ sudo ./bazel-bin/admin/install/install
Password:
2017/11/25 19:33:24 bazel-bin/admin/install/install.runfiles/sandboxfs/cmd/sandboxfs/sandboxfs -> /usr/local/bin/sandboxfs (mode -rwxr-xr-x)
2017/11/25 19:33:24 bazel-bin/admin/install/install.runfiles/sandboxfs/AUTHORS -> /usr/local/share/doc/sandboxfs/AUTHORS (mode -rw-r--r--)
2017/11/25 19:33:24 bazel-bin/admin/install/install.runfiles/sandboxfs/CONTRIBUTING.md -> /usr/local/share/doc/sandboxfs/CONTRIBUTING.md (mode -rw-r--r--)
2017/11/25 19:33:24 bazel-bin/admin/install/install.runfiles/sandboxfs/CONTRIBUTORS -> /usr/local/share/doc/sandboxfs/CONTRIBUTORS (mode -rw-r--r--)
2017/11/25 19:33:24 bazel-bin/admin/install/install.runfiles/sandboxfs/INSTALL.md -> /usr/local/share/doc/sandboxfs/INSTALL.md (mode -rw-r--r--)
2017/11/25 19:33:24 bazel-bin/admin/install/install.runfiles/sandboxfs/LICENSE -> /usr/local/share/doc/sandboxfs/LICENSE (mode -rw-r--r--)
2017/11/25 19:33:24 bazel-bin/admin/install/install.runfiles/sandboxfs/NEWS.md -> /usr/local/share/doc/sandboxfs/NEWS.md (mode -rw-r--r--)
2017/11/25 19:33:24 bazel-bin/admin/install/install.runfiles/sandboxfs/README.md -> /usr/local/share/doc/sandboxfs/README.md (mode -rw-r--r--)
2017/11/25 19:33:24 bazel-bin/admin/install/install.runfiles/sandboxfs/cmd/sandboxfs/sandboxfs.1 -> /usr/local/share/man/man1/sandboxfs.1 (mode -rw-r--r--)
```

Comment (rupertks) [reply INSTALL.md]: Works in Linux too (that was OS X above)

**Comment by rupertks** [comment]
The install works for me now, but I couldn't actually run sandboxfs afterwards - raised https://github.com/bazelbuild/sandboxfs/issues/8

File: admin/install/install.go [hunk 2]
```diff
@@ -0,0 +1,121 @@
...
+	{"LICENSE", "share/doc/sandboxfs", dataPerm},
+	{"NEWS.md", "share/doc/sandboxfs", dataPerm},
+	{"README.md", "share/doc/sandboxfs", dataPerm},
+
+	{"cmd/sandboxfs/sandboxfs.1", "share/man/man1", dataPerm},
+}
+
+// doInstall performs the installation of all given data files into the prefix.
+func doInstall(sourceDir string, prefix string, data []installSpec) error {
+	for _, spec := range data {
```
rupertks: installData instead of data?

or filesToInstall ?

admin/install/install.go
```diff
@@ -0,0 +1,121 @@
...
+	// targetDir indicates the relative path from the prefix where the file will be installed.
+	// The name of the installed file matches the name of the source file.
+	targetDir string
+
+	// mode indicates the desired permissions of the installed file.
+	mode os.FileMode
+}
+
+// installData specifies the files to be installed by this tool.
+var installData = []installSpec{
```
rupertks: Does installData need to be global?

configure
```diff
@@ -0,0 +1,85 @@
...
+# this were a Go program and we wanted to tell users to invoke it with "bazel
+# run" as we do for all other tools, bazel-gopath would get stuck because Bazel
+# can only run one command at a time.  It's easier for now to just keep this as
+# a shell-script so that it's impossible to run it from within Bazel.
+#
+# TODO(jmmv): Convert to a Go program once bazel-gopath is fixed to not need
+# to run Bazel.  (E.g. it could be made to read query outputs from a dependency
+# and then have a Bazel genquery rule to supply such data as a dependency.)
+
+set -e
```
rupertks: set -euo pipefail?

admin/install/install.go
jmmv: It doesn't *have* to be, but I considered this to be the program's data and thus felt better represented as a constant. (It cannot be a constant, though, because constants in Go can only be of primitive types.)

admin/install/install.go
jmmv: `installData` might be nice, but I'm wary of shadowing global symbols because that can lead to more confusion. `filesToInstall` sounds nice too, but then I'd also want to rename the global to match this new name so the same issue would arise.

I feel that `data` here is not too bad given that the type definition is immediately before use, the type is descriptive enough, and this is a private function (so the text wouldn't end up in godoc).

Don't feel too strongly about this, so if you have a better suggestion to avoid the shadowing issue...

jmmv:
I've fixed up the last commit as promised into the corresponding commit and also done a couple of changes to the `configure` script (homogenized global variable naming and added `readonly` qualifiers to them).

These changes require reapproval for merge. The only differences should be in the `configure` script.

Thanks for your patience! I'll try to stay to single-commit pull requests in the future so that things are easier to deal with in GitHub.

admin/install/install.go
rupertks: Personally I would rename both to filesToInstall and move the global version of filesToInstall into main

configure
rupertks: (Just bumping this because GitHub says the previous review was dismissed)

Is it not safe to use set -euo pipefail for the configure script?

rupertks:
Just added a few more comments, although they don't show up at the bottom of the timeline on this page. GitHub seems... not great for reviews :)

configure
jmmv: Wut? I'm sure I typed a reply here!

`set -u`: This would be nice... but the ancient bash (which is `/bin/sh`) shipped on macOS has bugs that cause it to complain on references to `"${@}"` when there are no arguments — and references to empty `"${@}"` should be valid. This makes the call to main, and the argument processing, to fail when this is enabled.

`set -o pipefail`: This is a bash-only feature and I don't want to rely on bash here.

admin/install/install.go
jmmv: Done.

PR #10, opened 2017-11-29
Implement a --version flag

jmmv:
Add a version number to sandboxfs and implement a --version flag to dump
it to stdout.  This will be used by Bazel to determine which protocol to
use to communicate with the sandboxfs daemon, allowing us to change the
command-line interface in the future.

Personally, I don't like --version being a flag: given that we have a
subcommand-based interface, this should really be a "version" command.
However, we already have --help so --version doesn't hurt much more.
And I have ideas to kill the subcommand-based interface, which would
homogenize things.

### DIFF
--- a/integration/cli_test.go
+++ b/integration/cli_test.go
@@ -24,7 +24,7 @@ import (
 
 var (
 	// versionPattern contains a pattern to match the output of sandboxfs --version.
-	versionPattern = `sandboxfs [0-9]+.[0-9]+`
+	versionPattern = `sandboxfs [0-9]+\.[0-9]+`
 )
 
 func TestCli_Help(t *testing.T) {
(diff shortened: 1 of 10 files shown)


jmmv:
Meh, was trickier than I thought to make the tests pass. But they do now!

All the commits in this pull request ought to be squashed at merge time so the description of all commits but the first do not matter.

integration/cli_test.go
```diff
@@ -16,11 +16,17 @@ package integration
...
 	"fmt"
+	"os"
 	"testing"
 
 	"github.com/bazelbuild/sandboxfs/integration/utils"
 )
 
+var (
+	// versionPattern contains a pattern to match the output of sandboxfs --version.
+	versionPattern = `sandboxfs [0-9]+.[0-9]+`
```
rupertks: Should this be ```\.``` ?

rupertks:
Personally I like --version working (and I wish bazel --version worked, too)

integration/cli_test.go
jmmv: Yes.

jmmv:
WRT Bazel, see https://github.com/bazelbuild/bazel/pull/3671

PR #11, opened 2017-12-21
Handle transient unmount errors at test teardown

jmmv:
It is possible, although rare, for a system component such as macOS's
Finder to interfere with the sandboxfs test instances we set up.  One
of the interferences we observe is that the Finder may access the
file system while we are running the test teardown routine, and thus
the unmount we issue fails due to a "resource busy" error.

Under those conditions, the only thing we can do is retry the unmount
operation because the fact that it failed is not our fault.

This should fix some very rare test flakiness I've observed on Travis
and also resolves some consistent flakiness I was observing in new
tests I'm adding, which left me puzzled for a long time.

### DIFF
--- a/integration/utils/exec.go
+++ b/integration/utils/exec.go
@@ -109,17 +109,20 @@ func RunAndWait(wantExitStatus int, arg ...string) (string, string, error) {
 	return wait(state, wantExitStatus)
 }
 
-// waitForFile waits until a given file exists and is accessible by the given user with a timeout
-// of deadlineSeconds.  Returns true if the file was found before the timer expired, or false
-// otherwise.  The user may be nil, in which case the current user is assumed.
-func waitForFile(cookie string, deadlineSeconds int, user *UnixUser) bool {
-	for tries := 0; tries < deadlineSeconds; tries++ {
-		if err := FileExistsAsUser(cookie, user); err == nil {
-			return true
+// retry runs the given action until either it succeeds or the given deadline expires.  If the
+// deadline expires, returns the last encountered error from the action.  Prints the given message
+// when an error is encountered.
+func retry(action func() error, message string, deadlineSeconds int) error {
+	var lastErr error
+	for tries := 0; tries < deadlineSeconds*10; tries++ {
+		lastErr = action()
+		if lastErr == nil {
+			return nil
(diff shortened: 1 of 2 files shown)


integration/utils/exec.go
```diff
@@ -410,12 +414,24 @@ func mountSetupFull(t *testing.T, stdout io.Writer, stderr io.Writer, user *Unix
 // If tests wish to control the shutdown of the sandboxfs process, they can do so, but then they
 // must set s.Cmd to nil to tell TearDown to not clean up the process a second time.  The same
 // applies to s.Stdin.
-func (s *MountState) TearDown(t *testing.T) {
+//
+// If tests wish to check if TearDown returned an error, they can do so by avoiding the recommended
+// use of "defer".  Note, though, that such tests will only receive the first error encountered by
+// this function, and that the function will run to completion even if there were failures.
+func (s *MountState) TearDown(t *testing.T) error {
+	var finalErr error
```
rupertks: Call this firstErr instead? 

That seems to be what it is, and the comment above talks about "first error" too

jmmv: Done. "final" here meant "error to be returned in the end".

philwo:
> Under those conditions, the only thing we can do is retry the unmount
operation because the fact that it failed is not our fault.

What would happen if we accidentally left a process running in that directory? For example, if a test spawns a daemon that never exits. (process-wrapper cannot prevent this at the moment, there is no known mechanism on macOS to track daemonizing child processes)

jmmv:
Note that this fix is for our own tests, not a fix in sandboxfs itself.

If we leak a process from our own tests, then the various retries to unmount during teardown will fail and the test will abort. I think this is acceptable given that this is a bug on our side.

If this happens from Bazel where a process is left behind as you describe... well, the mount point will be left busy and Bazel won't be able to unmount it. That will need a different fix (to code that's not yet submitted!), and we can probably resolve that from Bazel's side. Note that BSDs (and thus macOS) have an "force" flag to the unmount system call / command, which causes any active file descriptors to be revoked. This should work for the case you are describing.

philwo:
Oh, sorry! Of course. In that case this is fine :)

PR #13, opened 2018-01-27
Use Bazel 0.9.0 and Go rules 0.9.0

jmmv:
As part of this change, rerun Gazelle to update all BUILD.bazel files
with new changes.
    
We also have to workaround a change in the paths of the built
Go binaries (https://github.com/bazelbuild/rules_go/issues/1239) in our
instructions and build scripts.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -27,7 +27,7 @@ To get started:
     build the final binary, and build the installation tool.
 
 1.  To copy all binary and data files to the default destination of
-    `/usr/local`, run `./bazel-bin/admin/install/install`.
+    `/usr/local`, run `./bazel-bin/admin/install/*/install`.
 
     *   You will (most likely) need superuser permissions to install
         under `/usr/local`, so run the previous command with `sudo`.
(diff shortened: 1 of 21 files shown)


jmmv:
The amount of time I wasted on this because of a change in the Go rules is insane (and I find it invalid anyway)... but I have no better answers.

jmmv:
It seems like the workaround for the issue quoted above is getting some traction; see https://github.com/bazelbuild/bazel-gazelle/issues/106. I've withdrawn the review request until I have more data, as if that happens, this PR will become much, much simpler.

rupertks:
Okay!

Let me know when I should look at this again.

jmmv:
OMG. This has been *awful* but I think it finally works. I've also spent some time breaking apart the huge change into three more-manageable pieces. The first two build up support to switch to a new rules version without actually changing that, and then the last change becomes "trivial". Haha, I only wish it had been. :'(

jmmv:
I moved a lot of the code in the first commit of the three in this PR upstream (https://github.com/bazelbuild/rules_go/pull/1331). This is now much simpler and should be ready for review.

PR #16, opened 2018-02-27
Drop the subcommand-based interface

jmmv:
Merge the static and dynamic commands into one, which allows us to drop the
subcommand-based interface.  There never was a good reason to split the
behavior: we can just say that a sandboxfs instance starts running with the
mappings provided on the command line and that it can later be reconfigured
at will via the input/output files.

I'm intentionally not changing the flag names nor the general behavior of
reconfigurations to keep this change simple: fixing callers of sandboxfs
(e.g. all the integration tests) is as simple as dropping the "static"
and "dynamic" subcommand names from the calls.  Moving onto a different
interface is desirable (gRPC could possibly be a nice fit) but, if done,
will come later.

This is a massive simplification that will allow for upcoming changes to
the way the reconfiguration protocol works.

### DIFF
--- a/cmd/sandboxfs/sandboxfs.1
+++ b/cmd/sandboxfs/sandboxfs.1
@@ -11,7 +11,7 @@
 .\" WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
 .\" License for the specific language governing permissions and limitations
 .\" under the License.
-.Dd January 21, 2018
+.Dd February 27, 2018
 .Dt SANDBOXFS 1
 .Os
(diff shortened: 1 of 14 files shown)


jmmv:
Suggested review order: start with the changes to the manual page to understand the new API (`sandboxfs.1`), then continue onto the changes to the code itself (`sandboxfs.go` and `sandbox.go`), and later look at the tests.

The changes to the tests are many but are for the most part mechanical: I simply did a search/replace to strip out `"static"` and `"dynamic"` from all invocations.

PR #17, opened 2018-03-02
Respect intermediate directory components in nested mappings

jmmv:
If the user maps /foo/bar/baz into the sandbox and /foo/bar already exists
on disk as a descendent of the mapping for root, we should respect both
/foo's and /foo/bar's writability properties.  Do this by transparenty
turning them into mappings to their underlying directories.

We have to do this because that's what the user may expect.  For example:
Bazel creates a directory tree first and then maps individual files as
read-only within that tree, but it expects the rest of the tree to remain
writable as configured on the root node.

This fixes a regression introduced by the unification of directory node
types in PR #14.

### DIFF
--- a/internal/sandbox/dir.go
+++ b/internal/sandbox/dir.go
@@ -123,7 +123,27 @@ func (d *Dir) LookupOrCreateDirs(components []string) *Dir {
 		}
 		return nil
 	}
-	child := newDirEmpty()
+
+	var child *Dir
+	if underlyingPath, isMapped := d.UnderlyingPath(); isMapped {
+		childPath := filepath.Join(underlyingPath, name)
+		fileInfo, err := os.Lstat(childPath)
+		if err == nil {
+			if fileInfo.IsDir() {
+				child = newDir(childPath, fileInfo, d.writable)
+			} else {
+				log.Printf("Mapping clobbers non-directory %s with a read-only directory", childPath)
+				child = newDirEmpty()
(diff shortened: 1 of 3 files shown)


internal/sandbox/dir.go
```diff
@@ -123,7 +123,27 @@ func (d *Dir) LookupOrCreateDirs(components []string) *Dir {
...
+		if err == nil {
+			if fileInfo.IsDir() {
+				child = newDir(childPath, fileInfo, d.writable)
+			} else {
+				log.Printf("Mapping clobbers non-directory %s with a read-only directory", childPath)
+				child = newDirEmpty()
+			}
+		} else {
+			if !os.IsNotExist(err) {
+				log.Printf("Mapping clobbers %s due to IO error: %v", childPath, err)
```
rupertks: Have you ever seen this mystery (IO?) error happen? Do you have some examples?

Why is it safe for sandboxfs to continue (making the mapping, etc) if it gets this error?

jmmv: Nope, haven't seen it happen, but it's not that mysterious. Invalid permissions on a path component could lead to an error that is different than "not exist".

The reason for just logging this and continuing is the same as for the case where we clobber a file or directory: if the user specified a mapping, they most likely want that mapping to appear in the sandbox regardless of what the underlying file system said. Plus it's easier to handle this way :P

PR #18, opened 2018-03-02
Split the reconfiguration protocol into individual map/unmap operations

jmmv:
The previous sandboxfs exposed a mechanism to change the whole file
system contents "at once" (though not atomically).  This was problematic
because sandboxfs couldn't make any assurances about items being
replaced.

What's worse is that the previous approach made it very hard for Bazel
to share a single sandboxfs instance for different actions.  We'd like
to have each action run in a subdirectory of sandboxfs, but changing
the views of these directories without interfering others was not
possible.  (Bazel could have maintained a full list of all the mappings
and then fed this whole list to sandboxfs for each action... but that's
costly, and due to our poor reconfiguration algorithm, this caused
interference for entries that were supposed to remain unchanged.)

To make this all simpler, change the reconfiguration protocol so that
the user feeds a series of ordered steps to map and unmap independent
entries.  For Bazel, this trivially allows using different top-level
subdirectories for each action.  For other users... this may not be
great, but we don't care for now.  If we want to offer full
reconfigurations at a later stage, we can do so again.

### DIFF
--- a/internal/sandbox/dir.go
+++ b/internal/sandbox/dir.go
@@ -129,23 +129,28 @@ func (d *Dir) LookupOrCreateDirs(components []string) *Dir {
 	return child.LookupOrCreateDirs(remainder)
 }
 
-// LookupOrFail traverses a set of components and returns the directory containing the final
+// LookupOrFail traverses a set of path components and returns the directory containing the final
 // component, or nil if not found.
+//
+// The list of components is the result of splitting a path into its components and the list cannot
+// be empty on entry. The empty list would correspond to "/", but we cannot handle it here: the
+// caller must deal with the root special case on its own.
 func (d *Dir) LookupOrFail(components []string) *Dir {
-	if len(components) == 0 {
-		return d
-	}
 	name := components[0]
 	remainder := components[1:]
 
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
-	if child, ok := d.children[name]; ok {
-		if dir, ok := child.(*Dir); ok {
-			return dir.LookupOrFail(remainder)
-		}
+	child, ok := d.children[name]
+	if !ok {
 		return nil
+	}
+	if len(remainder) == 0 {
+		return d
+	}
+	if dir, ok := child.(*Dir); ok {
+		return dir.LookupOrFail(remainder)
 	}
 	return nil
 }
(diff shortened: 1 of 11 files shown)


internal/sandbox/dir.go
```diff
@@ -129,6 +129,27 @@ func (d *Dir) LookupOrCreateDirs(components []string) *Dir {
 	return child.LookupOrCreateDirs(remainder)
 }
 
+// LookupOrFail traverses a set of components and returns the directory containing the final
+// component, or nil if not found.
+func (d *Dir) LookupOrFail(components []string) *Dir {
+	if len(components) == 0 {
```
rupertks: Is this still correct for the case where components is initially zero length (so no recursion happens). Should LookupOrFail return nil in that case, instead of returning d?

This function could also do with an example of what a component/components string look like.

Also, why the "OrFail" suffix... ?

internal/sandbox/dir.go
```diff
@@ -144,6 +165,28 @@ func (d *Dir) Map(name string, node Node) error {
 	return nil
 }
 
+// Unmap unregisters a directory entry and all of its descendents.
+//
+// This function is intendd to be used during reconfigurations to delete parts of the tree that
```
rupertks: // This function is intended

internal/sandbox/sandbox.go
```diff
@@ -172,40 +165,137 @@ func tokenizePath(path string) []string {
...
+// reconfigureUnmap applies a single Unmap step to the file system.
+//
+// Note that this only removes the leaf of the given mapping: nested entries created during a map
+// operation are left behind and have to be unmapped explicitly.  E.g. if /foo/bar/baz was mapped
+// on an empty file system, /foo/bar are two intermediate mappings that are not removed when
+// /foo/var/baz is unmapped.  This is for simplicity as we'd otherwise need to track whether a
+// mapping was explicitly created by the user or as a result of another mapping, and also to keep
+// the unmapping algorithm trivial.
+func (f *FS) reconfigureUnmap(server *fs.Server, mapping string) error {
+	components, err := tokenizeMapping(mapping)
```
rupertks: I know that Go encourages short names, but this is really too vague, I think

Can we rename components to pathComponents or to pathTokens? Something that makes it clearer it is a path, anyway.

PR #24, opened 2018-07-26
Add a trivial root directory implementation

jmmv:
This change adds the foundations of the in-memory representation of the
contents of a sandboxfs instance.  In particular, a Node trait that will
handle all node-related operations and the basics of a Dir
implementation.

The implementation of the root directory does nothing interesting at
this point: it just exposes the . and .. entries and returns ENOENT for
any lookups, which is sufficient to introduce the basic file system
concepts.

Note that this does not yet allow us to enable more tests in
travis-build.sh, so the new code introduced by this change was purely
tested locally by mounting the file system and ensuring an "ls -a" shows
the dot entries and that accessing any file returns ENOENT.


jmmv:
I haven't merged the previous #23 yet because I'm waiting for @mhlopko's review, but I thought it'd be good to start discussing the next change already (hence why this PR is based off the `rust` branch, not `master`).

The design here tries to mimic the way the Go implementation works. I can share the full prototype of sandboxfs in Rust that I have lying around if it can help in providing more clarity on where I'm headed; these PRs are an attempt at cleaning that up step by step. Let me know.

Thanks in advance!

taralx:
Do you plan to support multi-user systems?

jmmv:
I'm not exactly sure what you want to get to...

I do not particularly want to support multiple users accessing the file system at once (our use cases are targeted at a single user running builds on top of sandboxfs), but we must support `allow_other` and mount file systems as such because of http://julio.meroh.net/2017/10/fighting-execs-sandboxfs-macos.html . So I guess the answer to your question is "yes"?

taralx:
Question is do you want to support the case of multiple users running builds concurrently. They'll each need a sandboxfs, and if sandboxfs is a system-managed daemon, then you might have trouble. If users can start & mount sandboxfs without root's help, then this is fine.

taralx:
Also the case of "user A logs in on console, builds, user B logs in remotely, builds".

jmmv:
Ah yes. Bazel is in charge of mounting one sandboxfs instance per build. Each instance is owned by that process and thus runs with the user's credentials.

src/lib.rs
```diff
@@ -13,24 +13,103 @@
...
+            let now = time::get_time();
+            let uid = unsafe { libc::getuid() } as u32;
+            let gid = unsafe { libc::getgid() } as u32;
+            nodes::Dir::new_empty(inode, inode, now, uid, gid)
+        };
+
+        let mut nodes = HashMap::new();
+        nodes.insert(root.inode(), root);
+        SandboxFS {
+            nodes: Arc::from(Mutex::from(nodes)),
```
taralx: Generally one uses ::new instead of ::from. Is there a reason for the latter?

src/nodes/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+mod dir;
+pub use self::dir::Dir;
+
+/// Abstract representation of a file system node.
+///
+/// Due to the way nodes and node operations are represented in the kernel, this trait exposes a
+/// collection of methods that do not all make sense for all possible node types: some methods will
+/// only make sense for directories and others will only make sense for regular files, for example.
+/// These conflicting methods come with a default implementation that panics.
```
taralx: Dislike panics. Can we avoid the panics with type magic? Or at least have the default impl return some kind of reasonable error (with log if possible).

src/lib.rs
```diff
@@ -13,24 +13,103 @@
...
+        }
     }
 }
 
 impl fuse::Filesystem for SandboxFS {
+    fn getattr(&mut self, _req: &fuse::Request, inode: u64, reply: fuse::ReplyAttr) {
+        let node = self.find_node(inode);
+        match node.getattr() {
+            Ok(attr) => reply.attr(&TTL, &attr),
+            Err(e) => reply.error(e.raw_os_error().unwrap()),
```
taralx: Very uncomfortable with this unwrap. It constrains the nodes heavily.

src/lib.rs
```diff
@@ -13,24 +13,103 @@
...
+            Ok((node, attr)) => {
+                {
+                    let mut nodes = self.nodes.lock().unwrap();
+                    if !nodes.contains_key(&node.inode()) {
+                        nodes.insert(node.inode(), node);
+                    }
+                }
+                reply.entry(&TTL, &attr, 0);
+            },
+            Err(e) => reply.error(e.raw_os_error().unwrap()),
```
taralx: Also here.

src/lib.rs
```diff
@@ -13,24 +13,103 @@
...
+        }
+    }
+
+    fn lookup(&mut self, _req: &fuse::Request, parent: u64, name: &OsStr, reply: fuse::ReplyEntry) {
+        let dir_node = self.find_node(parent);
+        match dir_node.lookup(name) {
+            Ok((node, attr)) => {
+                {
+                    let mut nodes = self.nodes.lock().unwrap();
+                    if !nodes.contains_key(&node.inode()) {
```
taralx: This is more commonly written `nodes.entry(nodes.inode()).or_insert(node);`.

src/lib.rs
```diff
@@ -13,24 +13,103 @@
...
+    fn readdir(&mut self, _req: &fuse::Request, inode: u64, _handle: u64, offset: i64,
+               mut reply: fuse::ReplyDirectory) {
+        if offset == 0 {
+            let node = self.find_node(inode);
+            match node.readdir(&mut reply) {
+                Ok(()) => reply.ok(),
+                Err(e) => reply.error(e.raw_os_error().unwrap()),
+            }
+        } else {
+            assert!(offset > 0, "Do not know what to do with a negative offset");
```
taralx: More reasonable to return an error than to die here. If we had logging I'd suggest logging an error as well.

src/nodes/dir.rs
```diff
@@ -0,0 +1,99 @@
...
+    ///
+    /// `time` is the timestamp to be used for all node times.
+    ///
+    /// `uid` and `gid` indicate the ownership details of the node.  For scaffold directories, these
+    /// should always match the values of the currently-running process -- but not necessarily if we
+    /// want to let users customize these via flags at some point.
+    ///
+    /// Scaffold directories are in-memory directories not backed by any on-disk contents.  These
+    /// are used to represent user-specified mappings
+    pub fn new_empty(inode: u64, parent: u64, time: Timespec, uid: u32, gid: u32) -> Arc<Node> {
```
taralx: Maybe this should be new_root? I feel like it would be safer to have a "mkdir" method to make child directories.

src/lib.rs
jmmv: Just me not knowing any better... changed now, but what's the difference? Just idiomatic?

src/lib.rs
jmmv: The `reply.error()` calls expect a libc error code (because that's the only thing the kernel understands in reply to one of these hooks) so there is not much more we can do if we don't get such detail...

Note that, in the Go version, we returned arbitrary errors from helper functions and, in the FUSE layer, used a "conversion" function that either extracted libc error codes from some error types or mapped "known" error types to libc error codes. I think that was a mistake: it's pretty hard to know what's going on regarding errors there because this function does a lot of magic. I wouldn't like to repeat this here.

One alternative I can think about is changing my helper functions to return a `Result<?, i32>` so that we push the responsibility of computing the libc error codes to the helpers.

What do you think? Or what other alternative do you have in mind? Remember: I'm very new to Rust so there may not be a good explanation for what I do :-P

src/lib.rs
jmmv: Done.

src/nodes/dir.rs
jmmv: This won't be used exclusively for the root directory though.

There are two variants of directory nodes: "mapped" directories, which correspond to a "real" directory in the underlying file system and which might have explicit overrides for certain entries; and "scaffold" directories, which are pure in-memory directories that initially start empty and gain entries (mapped or scaffold) as the user applies new mappings to them.

These two concepts were originally separate in the Go implementation as two different node types, but that caused a lot of problems to represent the root directory: during a "reconfiguration" (when we change the mappings the file system exposes without a remount), the type of the directories can change -- and the kernel (or rather, the FUSE library) doesn't like it if the type of the root directory changes. Hence why the two abstractions were merged. See https://github.com/jmmv/sandboxfs/commit/5b9f460b92853e814f1979786b5adf55ce3da912, https://github.com/jmmv/sandboxfs/commit/4d5042e21bea2d5c5808eb898b819e287c06dd5d, and https://github.com/jmmv/sandboxfs/commit/f9d56c7da24db26351eea3e6a1283c364519a9a5 for some context.

Actually, merging the two types brought in a lot of simplicity to the code: both directory types behave pretty much in the same manner and the only difference is whether one contains "real" entries or not. The restriction that the "root cannot change" doesn't apply in this Rust implementation because we are defining our own nodes and can do as we please, but I think it makes sense to keep them as a single entity for simplicity.

Thoughts for a better name welcome, but I think the context can help here.

src/nodes/mod.rs
jmmv: The reason I went for the panics (with this design) is because node implementations will need to be fully specified. For example: the `Symlink` node has to implement all unimplemented methods (like `inode`) and must only override the `readlink` method. If we end up calling any of the other methods with defaults (like `lookup`), that's a big problem because it makes no sense to do so on a symlink node. I'd rather know these things with a crash than to continue and cause potential corruption. That's why I don't think "sane" default implementations are a good thing here: they are bound to hide subtle design problems.

But I agree with you: they are ugly. I'd very much prefer some type magic to only call things when they are defined, but to me that seems 1. hard due to the dynamic nature of nodes, and 2. maybe slower? Inspecting types at run time sounds costly, at least in other languages. I'm afraid I don't know enough Rust to propose a better solution though. Any suggestions?

src/lib.rs
jmmv: The value we get here as an offset is a kernel construct–the `readdir` system call that the user invokes has no concept of offset). We have to trust the kernel to send us valid data here, which means a positive or zero offset.

If we get a negative value, it either implies there is a bug in the kernel or, more likely, that we haven't thought about some case that has to be handled. Either way, we don't know why that is, which means our mental model of this function doesn't match reality. I'd rather crash in those situations than to proceed with a potential inconsistency...

jmmv:
Alright, merged the base change and rebased this one on top of the new master. Hopefully this didn't mess the review too much...

jmmv:
@taralx Any feedback on my responses? If at any point you get tired of these reviews, please let me know. I can find somebody else, but so far your expert input seems very valuable.

src/nodes/mod.rs
taralx: I don't have enough experience with rust types to tell you if type magic will do here. You might want to inquire on the discourse server to see if more experienced people have some ideas.

src/nodes/dir.rs
taralx: I'm not suggesting separate types, just separate methods. `parent: u64` is just asking for a referential integrity violation. Having "it's a root directory or you created it from another node" makes more sense. I suppose you could use `parent: Option<Node>` or similar, but it makes more sense to me to have separate `make_root` and `Node.mkdir` functions.

src/lib.rs
taralx: My concern is that you don't enforce the existence of a `raw_os_error` in the type system, so there's an ugly `unwrap` that is in the "should not happen" category. Honestly, `io::Error` is a horrible type and should be avoided in general. I suggest having a custom error type that focuses on appropriate libc error codes. You can have specific handling in there for the lack of a `raw_os_error` and avoid making these functions more complex.

src/lib.rs
taralx: `from` implementations sometimes have coercion semantics that you might not want to invoke. `new` is always direct construction without any coercion.

src/lib.rs
jmmv: I see; that's a good idea. I've added a `KernelError` type to represent these though I'm not sure about the names I've used in the type and methods. There is no conversion code from `io::Error` yet because it's not necessary, so it'll come in a later change.

src/nodes/dir.rs
jmmv: Oh I see, good point. OK, changed to `new_root` and simplified the API for now. Once the ability to create subdirectories/map entries exists, I'll figure out how to refactor this code to reuse whatever is possible.

jmmv:
Did a couple of changes as you suggested. OMG, I hate GitHub reviews. Hopefully you can easily figure out where my replies are -- or otherwise you can see the changes in the commit I just pushed.

Thanks for going through this again. Hope you enjoyed the conference!

taralx:
LGTM

src/lib.rs
```diff
@@ -12,25 +12,104 @@
...
+    fn readdir(&mut self, _req: &fuse::Request, inode: u64, _handle: u64, offset: i64,
+               mut reply: fuse::ReplyDirectory) {
+        if offset == 0 {
+            let node = self.find_node(inode);
+            match node.readdir(&mut reply) {
+                Ok(()) => reply.ok(),
+                Err(e) => reply.error(e.errno()),
+            }
+        } else {
+            assert!(offset > 0, "Do not know what to do with a negative offset");
```
kevincox: Does it make sense to use a `debug_assert` or to reply with an error?

src/nodes/mod.rs
kevincox: Why can't we just return an error to fuse? Then the error can be propagated back to the user without crashing the process. For example `readdir` can return `ENOTDIR` by default.

src/nodes/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// Constructs a new error given a raw errno code.
+    fn from_errno(errno: i32) -> KernelError {
+        KernelError { errno }
+    }
+
+    /// Obtains the errno code contained in this error, which can be fed back into the kernel.
+    pub fn errno(&self) -> i32 {
+        return self.errno;
+    }
+}
```
kevincox: Does it make sense to use [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html)? It supports going from/to errno (on unix).

src/nodes/dir.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// `uid` and `gid` indicate the ownership details of the node.  These should always match the
+    /// values of the currently-running process -- but not necessarily if we want to let users
+    /// customize these via flags at some point.
+    pub fn new_root(time: Timespec, uid: u32, gid: u32) -> Arc<Node> {
+        let inode = fuse::FUSE_ROOT_ID;
+
+        let attr = fuse::FileAttr {
+            ino: inode,
+            kind: fuse::FileType::Directory,
+            nlink: 2,  // "." entry plus whichever initial named node points at this.
```
kevincox: It's worth noting that some tools such as GNU find assume that `nlinks == 2` means that there are no subdirectories since each subdirectory will contain an extra link for `..`. It might be more of a bother then it is worth but even setting this value to 3 may avoid these tools misbehaving.

PR #28, opened 2018-09-11
Implement getattr on directories

jmmv:
This change starts to add support for directory nodes backed by other
directories in the file system, but only implements getattr on them so
far.  The reason is that the logic to manipulate stat data is complex
enough on its own so submitting it as a separate change will help with
the review process.  As a result, this change is pretty much useless
on its own.

I have tested that this works by mapping an arbitrary directory to the
root of the sandbox and ensuring that "stat" on it yields the desired
data.  We'll be able to enable more comprehensive integration tests
once the implementation of the directory is more complete.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,9 +76,9 @@ struct SandboxFS {
 
 impl SandboxFS {
     /// Creates a new `SandboxFS` instance.
-    fn new(mappings: &Vec<Mapping>) -> io::Result<SandboxFS> {
+    fn new(mappings: &[Mapping]) -> io::Result<SandboxFS> {
         let root = {
-            if mappings.len() == 0 {
+            if mappings.is_empty() {
                 let now = time::get_time();
                 let uid = unsafe { libc::getuid() } as u32;
                 let gid = unsafe { libc::getgid() } as u32;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,10 +16,12 @@
 extern crate fuse;
 extern crate libc;
 #[macro_use] extern crate log;
+#[cfg(test)] extern crate tempdir;
 extern crate time;
 
 use std::collections::HashMap;
 use std::ffi::OsStr;
+use std::fs;
 use std::io;
 use std::path::{Path, PathBuf};
 use std::result::Result;
@@ -74,19 +76,36 @@ struct SandboxFS {
 
 impl SandboxFS {
     /// Creates a new `SandboxFS` instance.
-    fn new() -> SandboxFS {
+    fn new(mappings: &Vec<Mapping>) -> io::Result<SandboxFS> {
         let root = {
-            let now = time::get_time();
-            let uid = unsafe { libc::getuid() } as u32;
-            let gid = unsafe { libc::getgid() } as u32;
-            nodes::Dir::new_root(now, uid, gid)
+            if mappings.len() == 0 {
+                let now = time::get_time();
+                let uid = unsafe { libc::getuid() } as u32;
+                let gid = unsafe { libc::getgid() } as u32;
+                nodes::Dir::new_root(now, uid, gid)
+            } else if mappings.len() == 1 {
+                if mappings[0].path != Path::new("/") {
+                    panic!("Unimplemented; only support a single mapping for the root directory");
+                }
+                let fs_attr = fs::symlink_metadata(&mappings[0].underlying_path)?;
+                if !fs_attr.is_dir() {
+                    warn!("Path {:?} is not a directory; got {:?}", &mappings[0].underlying_path,
+                        &fs_attr);
+                    return Err(io::Error::from_raw_os_error(libc::EIO));
+                }
+                nodes::Dir::new_mapped(fuse::FUSE_ROOT_ID, &mappings[0].underlying_path, &fs_attr)
+            } else {
+                panic!("Unimplemented; only support zero or one mappings so far");
+            }
         };
 
         let mut nodes = HashMap::new();
+        assert_eq!(fuse::FUSE_ROOT_ID, root.inode());
         nodes.insert(root.inode(), root);
-        SandboxFS {
+
+        Ok(SandboxFS {
             nodes: Arc::from(Mutex::from(nodes)),
-        }
+        })
     }
 
     /// Gets a node given its `inode`.
@@ -154,11 +173,7 @@ pub fn mount(mount_point: &Path, mappings: &[Mapping]) -> io::Result<()> {
         .iter()
         .map(|o| o.as_ref())
         .collect::<Vec<&OsStr>>();
-
-    // TODO(jmmv): Handle the list of mappings by passing it into the file system.
-    let _mappings = mappings;
-
-    let fs = SandboxFS::new();
+    let fs = SandboxFS::new(mappings)?;
     info!("Mounting file system onto {:?}", mount_point);
     fuse::mount(fs, &mount_point, &options)
         .map_err(|e| io::Error::new(e.kind(), format!("mount on {:?} failed: {}", mount_point, e)))
--- a/src/nodes/conv.rs
+++ b/src/nodes/conv.rs
@@ -179,13 +179,6 @@ mod tests {
         assert_eq!(BAD_TIME, timespec);
     }
 
-    /// Sets the permissions of a file to exactly those given.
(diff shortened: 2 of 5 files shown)


src/nodes/conv.rs
```diff
@@ -0,0 +1,384 @@
...
+
+/// Converts a system time as represented in a fs::Metadata object to a Timespec.
+///
+/// `path` is the file from which the timestamp was originally extracted and `name` represents the
+/// metadata field the timestamp corresponds to; both are used for debugging purposes only.
+///
+/// If the given system time is missing, or if it is invalid, logs a warning and returns a fixed
+/// time.  It is reasonable for these details to be missing because the backing file systems do
+/// not always implement all possible file timestamps.
+fn system_time_to_timespec(path: &Path, name: &str, time: &io::Result<SystemTime>) -> Timespec {
```
taralx: It's unusual to take a `Result` as a parameter. Perhaps the coercion to an `Option` should be done by the caller? (There's a method for that.)

src/nodes/conv.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    // TODO(jmmv): Using the underlying ctimes is slightly wrong because the ctimes track changes
+    // to the inodes.  In most cases, operations that flow via sandboxfs will affect the underlying
+    // ctime and propagate through here, which is fine, but other operations are purely in-memory.
+    // To properly handle those cases, we should have our own ctime handling.
+    let ctime = Timespec { sec: attr.ctime(), nsec: attr.ctime_nsec() as i32 };
+
+    let perm = match attr.permissions().mode() {
+        // TODO(https://github.com/rust-lang/rust/issues/51577): Drop :: prefix.
+        mode if mode > ::std::u16::MAX as u32 => {
```
taralx: Does this not work as `u16::MAX`?

src/nodes/conv.rs
```diff
@@ -0,0 +1,384 @@
...
+            &Err(io::Error::from_raw_os_error(libc::ENOENT)));
+        assert_eq!(BAD_TIME, timespec);
+    }
+
+    /// Sets the permissions of a file to exactly those given, bypassing the umask.
+    fn chmod_exactly(path: &Path, perm: libc::mode_t) {
+        let path = path.as_os_str().to_str().unwrap().as_bytes();
+        let path = CString::new(path).unwrap();
+        assert_eq!(0, unsafe {
+            let old_umask = libc::umask(0);
```
taralx: Uh, what? Since when does chmod pay any attention to umask?

src/lib.rs
```diff
@@ -74,19 +76,36 @@ struct SandboxFS {
...
-    fn new() -> SandboxFS {
+    fn new(mappings: &Vec<Mapping>) -> io::Result<SandboxFS> {
         let root = {
-            let now = time::get_time();
-            let uid = unsafe { libc::getuid() } as u32;
-            let gid = unsafe { libc::getgid() } as u32;
-            nodes::Dir::new_root(now, uid, gid)
+            if mappings.len() == 0 {
+                let now = time::get_time();
+                let uid = unsafe { libc::getuid() } as u32;
```
taralx: Consider bringing in a crate like `nix` to avoid the unsafe use here.

src/nodes/mod.rs
```diff
@@ -38,6 +40,12 @@ impl KernelError {
     }
 }
 
+impl From<io::Error> for KernelError {
+    fn from(e: io::Error) -> Self {
+        KernelError::from_errno(e.raw_os_error().unwrap())
```
taralx: Reminder that this may panic.

src/lib.rs
jmmv: I gave this a try... but `nix`'s `getpid` returns a `Uid` and I don't see a way to cast that back to a raw integer number, which I need later. Am I missing something obvious?

Also, while `nix` contains some useful functions that would avoid me having to make my own (like `mkfifo` in the `conv` tests), that's about it so far. I did not see replacements for all other functions I need to call via libc. In other words: I don't think using `nix` *right now* is a good idea because I cannot eliminate all references to `libc`; prefer to keep a single pattern.

src/nodes/conv.rs
jmmv: I see `ok()` and `err()`, both which return options for each -- but note that here I'm accessing both values (not just the OK value) because I want to log the error. Given that this is an internal helper function I wouldn't worry too much about this signature. If it has to become public at some point (it might), I'll change it.

src/nodes/conv.rs
jmmv: Nope:

```error[E0599]: no associated item named `MAX` found for type `u16` in the current scope
   --> src/nodes/conv.rs:111:24
    |
111 |         mode if mode > u16::MAX as u32 => {
    |                        ^^^^^^^^ associated item not found in `u16`
```

src/nodes/conv.rs
jmmv: Uh, what? Exactly. I have no clue what I was thinking. (Well, I know what led me to this stupidity, but still.) Removed.

src/nodes/mod.rs
jmmv: Handled now.

src/lib.rs
```diff
@@ -74,19 +76,36 @@ struct SandboxFS {
...
+                nodes::Dir::new_root(now, uid, gid)
+            } else if mappings.len() == 1 {
+                if mappings[0].path != Path::new("/") {
+                    panic!("Unimplemented; only support a single mapping for the root directory");
+                }
+                let fs_attr = fs::symlink_metadata(&mappings[0].underlying_path)?;
+                if !fs_attr.is_dir() {
+                    warn!("Path {:?} is not a directory; got {:?}", &mappings[0].underlying_path,
+                        &fs_attr);
+                    return Err(io::Error::from_raw_os_error(libc::EIO));
```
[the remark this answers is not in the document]
jmmv: This whole `SandboxFS::new` method is still a hack to cope with the incomplete implementation of mappings. Once I implement the handling for all mappings (which first needs full node implementations), this will have to change. Wouldn't worry too much yet...

src/lib.rs
jmmv: I've filed https://github.com/nix-rust/nix/issues/943 about the UID/GID issue. I've also filed a PR to implement `chmod` and was told there is a way to do this natively within Rust. I'll continue investigating what will it take to avoid calling directly into libc for this project. But for now I'd rather stick to libc until I can supply a PR that kills all usages in favor of nix.

src/nodes/dir.rs
```diff
@@ -16,7 +16,10 @@ extern crate fuse;
 extern crate libc;
 extern crate time;
 
+use super::conv;
```
taralx: Is there a way to make this crate-relative instead of module-relative?

src/nodes/dir.rs
```diff
@@ -78,7 +110,22 @@ impl Node for Dir {
...
-        let state = self.state.lock().unwrap();
+        let mut state = self.state.lock().unwrap();
+
+        let new_attr = match state.underlying_path.as_ref() {
+            Some(path) => {
+                let fs_attr = fs::symlink_metadata(path)?;
+                if !fs_attr.is_dir() {
+                    warn!("Path {:?} backing a directory node is no longer a directory; got {:?}",
+                          state.underlying_path, &fs_attr);
+                    return Err(KernelError::from_errno(libc::EIO));
```
taralx: One could maybe use ENOTDIR here, but EIO is fine.

src/lib.rs
```diff
@@ -74,19 +76,36 @@ struct SandboxFS {
 
 impl SandboxFS {
     /// Creates a new `SandboxFS` instance.
-    fn new() -> SandboxFS {
+    fn new(mappings: &Vec<Mapping>) -> io::Result<SandboxFS> {
```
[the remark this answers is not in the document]
jmmv: Done. Luckily Clippy also caught it after I rebased the PR :)

src/nodes/dir.rs
jmmv: Probably. For now I've reordered this chunk alphabetically, which collates this new `use` with other crate-relative `use`s. I notice there are several more instances of this already in other files, so I'd rather fix it in one go in a separate change.

jmmv:
Appeased Clippy warnings after rebasing this PR on top of master. Will merge once tests pass.

PR #29, opened 2018-09-16
Lint improvements

jmmv:
Misc improvements to the lint tool. You'll notice some similarities to code you've recently seen... ;-P

We should be doing one PR per commit to appease GitHub, but given that the two commits here are trivial, I couldn't bother to do that. Will "rebase and merge" them onto master when approved, not squash them.

### DIFF
--- a/admin/lint/checks.go
+++ b/admin/lint/checks.go
@@ -15,8 +15,10 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


jin:
> We should be doing one PR per commit to appease GitHub

Could you elaborate on that? I've been doing PRs with multiple commits and that has never really been an issue during reviews.

admin/lint/checks.go
```diff
@@ -46,6 +48,38 @@ func checkLicense(file string) error {
...
+// checkNoTabs checks if the given file contains any tabs as indentation and, if it does, returns
+// an error.
+func checkNoTabs(file string) error {
+	input, err := os.OpenFile(file, os.O_RDONLY, 0)
+	if err != nil {
+		return fmt.Errorf("failed to open %s for read: %v", file, err)
+	}
+	defer input.Close()
+
+	preg := regexp.MustCompile(`^ *\t`)
```
jin: what's `preg`? pattern regexp?

jmmv:
PRs with multiple commits work fine as you say, but GitHub's review system is suboptimal for that. I tried it for large patch sets and it was a mess. Especially if you fold any review comments into the individual commits, as the force push required for them messes up the review threads. (Of course, you'd save the rebase until the reviews are done, but that means any commits to address comments are harder to incorporate into the original commits.)

There was a discussion between @philwo and I somewhere where we discussed this issue in Bazel and the compromise of sticking to a PR per commit... but I cannot find it now.

PR #30, opened 2018-09-16
Clippy

jmmv:
Miscellaneous changes to enable Clippy and fix Rust warnings.

Not bothering to send separate PRs for each commit because they are trivial. Will "rebase and merge" once approved, not squash.


src/nodes/dir.rs
```diff
@@ -57,8 +57,8 @@ impl Dir {
             ctime: time,
             crtime: time,
             perm: 0o555 as u16,  // Scaffold directories cannot be mutated by the user.
-            uid: uid,
-            gid: gid,
+            uid,
```
taralx: I don't like this recommendation.

PR #31, opened 2018-09-22
Simplify and generalize some test code

jmmv:
A couple of commits that will be rebased on top of master and fast-forwarded once approved.

One is to remove a lot of custom code thanks to an upstream change (still waiting to be merged, but didn't want to hold getting this review out).

Another one is to generalize some test boilerplate so I can reuse it in an upcoming change. The upcoming change will add the ability to instantiate `Node`s of various types and I need to ensure that such logic can deal with all possible file types as well, hence `AllFileTypes` being generic is a good thing.


jmmv:
Thanks for the quick review!

PR #32, opened 2018-09-22
Add a cache for nodes and barebones file/symlink nodes

jmmv:
This PR is a bit long, sorry! But a couple of things that should make the review easier: the new `file.rs` and `symlink.rs` are pretty much duplicates. I find this unfortunate but cannot think of a way to remove the duplication. Doesn't bother me much because it's trivial though, and these files will later grow to contain node-specific logic. Also, there is quite a bunch of new test code here that bloats this PR; the actual new logic is all within `lib.rs`.

There are a bunch of TODOs in this new code, but I'm just trying to replicate the exact same functionality in the Go version of this codebase. (Most of the comments and TODOs come from there, as well as the general design.) I'd rather not change the design much (other than for making it Rust-friendly) because it'll be easier to make this reimplementation work with the existing workloads that the Go one supports.

----- Commit description -----

This change adds a factory for nodes in the form of a cache.  The cache
tracks nodes that map to specific underlying files across different
reconfigurations so that we don't lose precious in-kernel state for
files that may be mapped and unmapped many times.  This is specifically
useful in the context of a Bazel build where an action may need to map
the C++ compiler, the following may not need it, but a next one may
need it again: if we don't maintain the node, the kernel will think the
compiler is a different file each time and will have to reload it from
disk.

To make the cache functional, this change also adds barebone
implementations for all other node types we will need, namely File and
Symlink, and adds a "writable" property to each node.

Note, however, that this code is pretty much useless as it is because
the only call we have now is to create the node for the mount point.
This will become useful in the next commit, where I'll implement
read-only mapped directories (i.e. their readdir and lookup operations).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,7 +83,7 @@ impl IdGenerator {
 
     /// Obtains a new identifier.
     pub fn next(&self) -> u64 {
-        let id = self.last_id.fetch_add(1, Ordering::SeqCst);
+        let id = self.last_id.fetch_add(1, Ordering::AcqRel);
         // TODO(https://github.com/rust-lang/rust/issues/51577): Drop :: prefix.
         if id >= ::std::u64::MAX as usize {
             panic!("Ran out of identifiers");
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,6 +26,7 @@ use std::io;
 use std::path::{Path, PathBuf};
 use std::result::Result;
 use std::sync::{Arc, Mutex};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use time::Timespec;
 
 mod nodes;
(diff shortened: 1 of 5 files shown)


jmmv:
Any ETA for this review? Thanks.

src/nodes/dir.rs
```diff
@@ -110,22 +113,18 @@ impl Node for Dir {
...
-                    warn!("Path {:?} backing a directory node is no longer a directory; got {:?}",
-                          state.underlying_path, &fs_attr);
-                    return Err(KernelError::from_errno(libc::EIO));
-                }
-                conv::attr_fs_to_fuse(path, self.inode, &fs_attr)
-            },
-            None => state.attr,
-        };
-        state.attr = new_attr;
+        let check_type = |t: fs::FileType| if t.is_dir() { Ok(()) } else { Err("directory") };
```
taralx: I believe you can leave off the type annotation here.

src/lib.rs
```diff
@@ -69,6 +70,104 @@ impl Mapping {
...
+
+impl IdGenerator {
+    /// Constructs a new generator that starts at the given value.
+    fn new(start_value: u64) -> Self {
+        IdGenerator { last_id: AtomicUsize::new(start_value as usize) }
+    }
+
+    /// Obtains a new identifier.
+    pub fn next(&self) -> u64 {
+        let id = self.last_id.fetch_add(1, Ordering::SeqCst);
```
taralx: AcqRel is sufficient here. SeqCst is super expensive on some architectures.

jmmv: Hadn't read the various options before. Seems reasonable; done.

PR #33, opened 2018-09-28
Implement read-only directories (lookup and readdir)

jmmv:
Starting to become functional, but not enough to enable significant tests yet. The next steps will be to implement read-only files, (which I'm not doing now because they require boilerplate to handle open file handles), symlinks, and nested mappings. With these two, we should be able to enable a bunch of integration tests, and possibly need to fix bugs.

Thanks again!


jmmv:
Withdrawing this PR in favor of another one with read-only implementations of all nodes. It's not as much code as I expected.

PR #34, opened 2018-10-01
Switch from libc to nix to avoid unsafe blocks

jmmv:
Had to add a couple of missing features to `nix` for this to be possible, but it's now!


taralx:
Oops, looks like you need some compat handlers for utimensat.

jmmv:
Thanks for the quick turnaround. Unfortunately, testing has discovered that the `utimesat` that I added to `nix` doesn't exist on macOS 10.12, which we need to run the integration tests in Travis (because they haven't figured out how to load the FUSE kext unattendedly in newer releases). I'll have to circle back there and add a simpler `utimes`, which we can use here.

PR #35, opened 2018-10-01
Implement read-only nodes

jmmv:
The three commits in this PR implement the read-only vnops needed for functional directories, symlinks, and files. Each commit is pretty much standalone and rather small, but to reduce turnaround times, I've combined them in a single review request. Hopefully that's not much trouble. Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -180,6 +180,9 @@ struct SandboxFS {
     /// Mapping of inode numbers to in-memory nodes that tracks all files known by sandboxfs.
     nodes: Arc<Mutex<HashMap<u64, Arc<nodes::Node>>>>,
 
+    /// Mapping of handle numbers to file open handles.
+    handles: Arc<Mutex<HashMap<u64, Arc<nodes::Handle>>>>,
+
     /// Cache of sandboxfs nodes indexed by their underlying path.
     cache: Arc<Cache>,
 }
@@ -218,6 +221,7 @@ impl SandboxFS {
         Ok(SandboxFS {
             ids: ids,
             nodes: Arc::from(Mutex::from(nodes)),
+            handles: Arc::from(Mutex::from(HashMap::new())),
             cache: Arc::from(cache),
         })
     }
@@ -230,10 +234,18 @@ impl SandboxFS {
     /// numbers we have previously told it about.
     fn find_node(&mut self, inode: u64) -> Arc<nodes::Node> {
         let nodes = self.nodes.lock().unwrap();
-        match nodes.get(&inode) {
-            Some(node) => node.clone(),
-            None => panic!("Kernel requested unknown inode {}", inode),
-        }
+        nodes.get(&inode).expect("Kernel requested unknown inode").clone()
+    }
+
+    /// Gets a handle given its identifier.
+    ///
+    /// We assume that the identifier is valid and that we have a known handle for it; otherwise,
+    /// we crash.  The rationale for this is that this function is always called on handles
+    /// requested by the kernel, and we can trust that the kernel will only ever ask us for handles
+    /// numbers we have previously told it about.
+    fn find_handle(&mut self, fh: u64) -> Arc<nodes::Handle> {
+        let handles = self.handles.lock().unwrap();
+        handles.get(&fh).expect("Kernel requested unknown handle").clone()
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -207,8 +217,12 @@ impl SandboxFS {
         assert_eq!(fuse::FUSE_ROOT_ID, root.inode());
         nodes.insert(root.inode(), root);
 
+        #[cfg_attr(feature = "cargo-clippy", allow(redundant_field_names))]
```
taralx: We might want to `#!` this at the crate level.

src/lib.rs
```diff
@@ -220,10 +234,18 @@ impl SandboxFS {
     /// numbers we have previously told it about.
     fn find_node(&mut self, inode: u64) -> Arc<nodes::Node> {
         let nodes = self.nodes.lock().unwrap();
-        match nodes.get(&inode) {
-            Some(node) => node.clone(),
-            None => panic!("Kernel requested unknown inode {}", inode),
-        }
+        nodes.get(&inode).expect("Kernel requested unknown inode").clone()
```
taralx: `expect()` crashes the process. I know this is a "should not happen", but I would prefer to log and return an error here. The caller can EIO the kernel request.

(also below)

src/nodes/dir.rs
```diff
@@ -129,15 +136,60 @@ impl Node for Dir {
...
+            return Ok((node.clone(), refreshed_attr))
+        }
+
+        let (child, attr) = {
+            let path = match state.underlying_path.as_ref() {
+                Some(underlying_path) => underlying_path.join(name),
+                None => return Err(KernelError::from_errno(libc::ENOENT)),
+            };
+            let fs_attr = fs::symlink_metadata(&path)?;
+            let node = cache.get_or_create(ids, &path, &fs_attr, self.writable);
```
taralx: Is the IdGenerator ever used separately from the Cache? Maybe it should be a part of the Cache.

src/nodes/file.rs
```diff
@@ -91,4 +99,25 @@ impl Node for File {
...
+                return Err(KernelError::from_errno(libc::EACCES));
+            }
+            if flags & libc::O_WRONLY != 0 {
+                options.read(false);
+            }
+            options.write(true);
+        }
+        options.custom_flags(flags);
+
+        let file = options.open(state.underlying_path.as_ref().unwrap())?;
```
taralx: How can you be sure underlying_path is Some here?

src/nodes/dir.rs
```diff
@@ -129,15 +136,60 @@ impl Node for Dir {
...
+            let attr = conv::attr_fs_to_fuse(path.as_path(), node.inode(), &fs_attr);
+            (node, attr)
+        };
+        state.children.insert(name.to_os_string(), child.clone());
+        Ok((child, attr))
     }
 
-    fn readdir(&self, reply: &mut fuse::ReplyDirectory) -> NodeResult<()> {
-        let state = self.state.lock().unwrap();
+    fn readdir(&self, ids: &IdGenerator, cache: &Cache, reply: &mut fuse::ReplyDirectory)
```
taralx: I'm not a fan of passing down the fuse-specific thing here. It's actually confusing -- this function only calls "add" but not "ok" or "error". IMO this function should return a Vec of directory entries, or, even better, an Iterator of directory entries (which lets us implement pagination later).

src/nodes/symlink.rs
```diff
@@ -82,4 +82,10 @@ impl Node for Symlink {
 
         Ok(state.attr)
     }
+
+    fn readlink(&self) -> NodeResult<PathBuf> {
+        let state = self.state.lock().unwrap();
+
+        Ok(fs::read_link(state.underlying_path.as_ref().unwrap())?)
```
taralx: also here

src/lib.rs
jmmv: Done. Added a first commit to do this.

src/lib.rs
jmmv: I still don't understand why you'd like to treat what seem to be logic errors gracefully. Yes, we could hit this condition due to a bug, but we could also hit it due to bad expectations on our side.

Either way, I think crashing is the best thing to do. What I'm concerned about is that, if we ever encounter this, we are in a situation where our view of inodes (or handles) doesn't match that of the kernel. Which implies two things: we might not find the requested inodes... but we may also be handing the kernel information for the wrong inode. The latter seems pretty bad: users could peek into files they are not allowed to access if that were the case so I think it makes sense to crash as soon as possible.

src/nodes/dir.rs
jmmv: Yes. It's used to generate handle identifiers which are not related to the cache, and possibly also generation numbers later on.

src/nodes/file.rs
jmmv: Oops, good catch. Files and directories needn't have an optional underlying path... but I was led the wrong path when trying to share the implementation of `getattr`. Removed optional now.

src/nodes/symlink.rs
jmmv: Same as for files. Shouldn't have been optional in the first place.

src/nodes/dir.rs
jmmv: Agreed in principle. However, the reason I went with the current approach is to avoid creating an ephemeral vector for the return value, only to later iterate over it and copy all data into the kernel reply. The other problem is that this vector ought to contain only the data that `reply.add()` later needs... which means having to define another struct just for this.

For now I've just updated the description of the method to be clear on what the expectations on the `reply` are. Thoughts?

src/nodes/dir.rs
```diff
@@ -119,10 +119,20 @@ impl Node for Dir {
     fn getattr(&self) -> NodeResult<fuse::FileAttr> {
         let mut state = self.state.lock().unwrap();
 
-        let check_type = |t: fs::FileType| if t.is_dir() { Ok(()) } else { Err("directory") };
-        if let Some(attr) = super::get_new_attr(
-            self.inode, state.underlying_path.as_ref(), check_type)? {
-            state.attr = attr;
+        let new_attr = match state.underlying_path.as_ref() {
```
taralx: This should use `Option.map`. Once NLL is available, you can directly assign instead.

jmmv: Interesting... but can it? Note the error handling within this match pattern. I don't think I can put that within the map lambda?

taralx: Curses, foiled again.

jmmv:
Thanks! Rebased on top of the `nix` branch that I just merged. Will merge this once tests pass.

Next up: multiple mappings.

PR #37, opened 2018-10-05
Miscellaneous test and error reporting changes

jmmv:
These are all in preparation for supporting multiple mappings, but given that the commit that introduces that feature is quite complex on its own, I think it makes sense to review this bag of smaller changes on its own.

(I think I finally understood how `failure` works here...)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -230,7 +230,7 @@ impl SandboxFS {
                 if !fs_attr.is_dir() {
                     warn!("Path {:?} is not a directory; got {:?}", &mappings[0].underlying_path,
                         &fs_attr);
-                    return Err(Error::from(io::Error::from_raw_os_error(errno::Errno::EIO as i32)));
+                    return Err(io::Error::from_raw_os_error(errno::Errno::EIO as i32).into());
                 }
                 cache.get_or_create(&ids, &mappings[0].underlying_path, &fs_attr,
                     mappings[0].writable)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,6 +21,7 @@ extern crate nix;
 #[cfg(test)] extern crate tempdir;
 extern crate time;
 
+use failure::{Error, ResultExt};
 use nix::{errno, unistd};
 use std::collections::HashMap;
 use std::ffi::OsStr;
@@ -189,7 +190,7 @@ struct SandboxFS {
 
 impl SandboxFS {
     /// Creates a new `SandboxFS` instance.
-    fn new(mappings: &[Mapping]) -> io::Result<SandboxFS> {
+    fn new(mappings: &[Mapping]) -> Result<SandboxFS, Error> {
         let ids = IdGenerator::new(fuse::FUSE_ROOT_ID);
         let cache = Cache::default();
 
@@ -205,7 +206,7 @@ impl SandboxFS {
                 if !fs_attr.is_dir() {
                     warn!("Path {:?} is not a directory; got {:?}", &mappings[0].underlying_path,
                         &fs_attr);
-                    return Err(io::Error::from_raw_os_error(errno::Errno::EIO as i32));
+                    return Err(Error::from(io::Error::from_raw_os_error(errno::Errno::EIO as i32)));
                 }
                 cache.get_or_create(&ids, &mappings[0].underlying_path, &fs_attr,
                     mappings[0].writable)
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -40,12 +40,21 @@ mod nodes;
...
-#[fail(display = "path {:?} is not absolute", path)]
-pub struct PathNotAbsoluteError {
-    /// The path that caused this error.
-    pub path: PathBuf,
+/// An error indicating that a mapping specification (coming from the command line or from a
+/// reconfiguration operation) is invalid.
+#[derive(Debug, Eq, Fail, PartialEq)]
+pub enum MappingError {
+    /// A path was required to be absolute but wasn't.
+    #[fail(display = "path {:?} is not absolute", path)]
```
taralx: `#[fail(display = "path {} is not absolute", path.display())]`

src/lib.rs
```diff
@@ -205,7 +206,7 @@ impl SandboxFS {
                 if !fs_attr.is_dir() {
                     warn!("Path {:?} is not a directory; got {:?}", &mappings[0].underlying_path,
                         &fs_attr);
-                    return Err(io::Error::from_raw_os_error(errno::Errno::EIO as i32));
+                    return Err(Error::from(io::Error::from_raw_os_error(errno::Errno::EIO as i32)));
```
taralx: You might be able to use `.into()` here.

src/lib.rs
```diff
@@ -337,15 +338,15 @@ impl fuse::Filesystem for SandboxFS {
...
+pub fn mount(mount_point: &Path, mappings: &[Mapping]) -> Result<(), Error> {
     let options = ["-o", "ro", "-o", "fsname=sandboxfs"]
         .iter()
         .map(|o| o.as_ref())
         .collect::<Vec<&OsStr>>();
     let fs = SandboxFS::new(mappings)?;
     info!("Mounting file system onto {:?}", mount_point);
-    fuse::mount(fs, &mount_point, &options)
-        .map_err(|e| io::Error::new(e.kind(), format!("mount on {:?} failed: {}", mount_point, e)))
+    fuse::mount(fs, &mount_point, &options).context(format!("mount on {:?} failed", mount_point))?;
```
taralx: FYI `context(...)` hides the underlying error.

src/lib.rs
jmmv: Too good to be true :(

From https://boats.gitlab.io/failure/derive-fail.html:

```
Note that unlike code that would appear in a method, this does not use something like self.code or self.message; it just uses the field names directly. This is because of a limitation in Rust's current attribute syntax. As a result, you can only interpolate fields through the derivation; you cannot perform method calls or use other arbitrary expressions.
```

so the call to `display()` doesn't work.

src/main.rs
```diff
@@ -126,6 +126,18 @@ fn safe_main(program: &str, args: &[String]) -> Result<(), Error> {
     Ok(())
 }
 
+/// Flattens all causes of an error into a single string.
```
taralx: This doesn't seem in line with failure's model. The top-level error is supposed to have enough information -- the remaining chain is for debugging or introspection.

jmmv: Yeah... I reached this approach from the long analysis in: https://github.com/rust-lang-nursery/failure/issues/189, which I understand to mean that annotating errors with some extra details is not something that failure can do well now.

Do you think that just "rewriting" the original IO error with the extra context, "by hand" as we did before with `map_err` is better?

taralx: Bleh. The alternative is a lot of complex and noisy `with_context` calls. I suppose this will do.

src/lib.rs
jmmv: I'm not sure about where you mean by "here"... I do not see an `into` in neither `io::Error` nor `failure::Error`) :-/

Also this line (and whole code block for that matter) will be gone with the upcoming mappings-related changes so I wouldn't worry too much. (Still interested in knowing what you meant though.)

src/main.rs
jmmv: By the way: where do you get that the rest of the error chain is only for debugging? I've been re-reading the failure manual and I do not get that impression, especially by what's explained in the *Context* section under https://boats.gitlab.io/failure/fail.html. But there doesn't seem to be explicit guidance.

src/lib.rs
jmmv: Resolving per the other comment thread on flattening causes, as this is directly derived from that.

src/lib.rs
taralx: Where there's a from, there's an into: https://doc.rust-lang.org/std/convert/trait.Into.html#generic-implementations

jmmv: Aha, I see. Also found https://doc.rust-lang.org/rust-by-example/conversion/from_into.html before I saw your link. Changed here and also in `main.rs`.

However, I do not see guidance on when to use `from` vs. `into`...

jmmv:
Given that this is a misc bag of changes, took the chance of pushing one extra commit to address the issue you brought up a while ago about `use` statements.

PR #38, opened 2018-10-11
Support multiple mappings

jmmv:
Implement support for when multiple mappings are passed via the command
line.

The diff below is against a bunch of not-yet-merged patches (the `misc` branch), but sending it out already for preliminary review. I'm not super-happy about the `map` algorithm, but attempts to simplify it haven't resulted in much nicer code.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,11 +22,9 @@ extern crate nix;
 extern crate time;
 
 use failure::{Error, ResultExt};
-use nix::{errno, unistd};
 use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::fs;
-use std::io;
 use std::os::unix::ffi::OsStrExt;
 use std::path::{Component, Path, PathBuf};
 use std::result::Result;
@@ -96,6 +94,12 @@ impl Mapping {
         }
 
         Ok(Mapping { path, underlying_path, writable })
+    }
+
+    /// Returns true if this is a mapping for the root directory.
+    fn is_root(&self) -> bool {
+        let mut components = self.path.components();
+        components.next() == Some(Component::RootDir) && components.next().is_none()
     }
 }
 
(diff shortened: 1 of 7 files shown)


jmmv:
Modified PR to point to master now.

src/nodes/dir.rs
```diff
@@ -16,14 +16,34 @@ extern crate fuse;
...
+use std::io;
+use std::path::{Component, Path, PathBuf};
 use std::sync::{Arc, Mutex};
-use time::Timespec;
+
+/// Takes the components of a path and returns the first normal component and the rest.
+///
+/// This assumes that the input path is normalized and that the very first component is a normal
+/// component as defined by `Component::Normal`.
+fn split_components<'a>(components: &'a [Component]) -> (&'a OsStr, &'a [Component<'a>]) {
```
taralx: One is `Component`, the other `Component<'a>`. Can you make them match? Can you remove all the `'a` specifiers, since they're all the same?

src/nodes/dir.rs
```diff
@@ -37,46 +57,43 @@ struct MutableDir {
...
+    /// Creates a new scaffold directory to represent an in-memory directory.
     ///
-    /// `uid` and `gid` indicate the ownership details of the node.  These should always match the
-    /// values of the currently-running process -- but not necessarily if we want to let users
-    /// customize these via flags at some point.
-    pub fn new_root(time: Timespec, uid: unistd::Uid, gid: unistd::Gid) -> Arc<Node> {
-        let inode = fuse::FUSE_ROOT_ID;
+    /// The directory's timestamps are all set to the current time, and the ownership is set to the
+    /// current user.
+    pub fn new_empty(inode: u64, parent: Option<&Node>) -> Arc<Node> {
```
taralx: You'll find it easier to test if you take `now` as an argument.

src/nodes/dir.rs
```diff
@@ -107,6 +124,42 @@ impl Dir {
...
+            Some(path) => {
+                let child_path = path.join(name);
+                match fs::symlink_metadata(&child_path) {
+                    Ok(fs_attr) => {
+                        if fs_attr.is_dir() {
+                            Dir::new_mapped(ids.next(), &child_path, &fs_attr, self.writable)
+                        } else {
+                            info!("Mapping clobbers non-directory {} with an immutable directory",
+                                child_path.display());
+                            Dir::new_empty(ids.next(), Some(self))
```
taralx: Can we factor out this line somehow? I suggest putting `return Dir::new_mapped(...` and then having the remaining cases fall through to the other case.

src/nodes/dir.rs
```diff
@@ -171,7 +264,12 @@ impl Node for Dir {
         reply.add(state.parent, 1, fuse::FileType::Directory, "..");
         let mut pos = 2;
 
-        // TODO(jmmv): Handle user-provided mappings before underlying entries.
+        for (name, dirent) in &state.children {
```
taralx: This section could use a comment. It took me multiple readings to understand why you can't do this in the loop below.

src/nodes/dir.rs
```diff
@@ -181,16 +279,26 @@ impl Node for Dir {
...
-                conv::filetype_fs_to_fuse(&path, fs_attr.file_type()), &name);
+            reply.add(child.inode(), pos, fs_type, &name);
             // Do the insertion into state.children after calling reply.add() to be able to move
             // the name into the key without having to copy it again.
-            state.children.insert(name, child.clone());
+            let dirent = Dirent {
+                node: child.clone(),
+                explicit_mapping: false,
+            };
+            state.children.insert(name, dirent);
```
taralx: Where do we remove deleted entries from `state.children`?

src/lib.rs
```diff
@@ -97,6 +95,12 @@ impl Mapping {
 
         Ok(Mapping { path, underlying_path, writable })
     }
+
+    /// Returns true if this is a mapping for the root directory.
+    fn is_root(&self) -> bool {
+        let mut components = self.path.components();
```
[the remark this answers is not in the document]
jmmv: Aha, I knew there had to be an easier way to test for this.

src/lib.rs
```diff
@@ -212,34 +216,44 @@ struct SandboxFS {
...
+            (nodes::Dir::new_mapped(ids.next(), &first.underlying_path, &fs_attr, first.writable),
+                &mappings[1..])
+        } else {
+            (nodes::Dir::new_empty(ids.next(), None), mappings)
+        }
+    };
+
+    for mapping in rest {
+        let components = mapping.path.components().collect::<Vec<_>>();
+        assert_eq!(Component::RootDir, components[0], "Paths in mappings are always absolute");
```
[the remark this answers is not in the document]
jmmv: Oh, I didn't know about this. Yes.

src/nodes/dir.rs
jmmv: Made them match. I had to add these as otherwise the compiler complains for both return values with errors like:

```
error[E0106]: missing lifetime specifier
  --> src/nodes/dir.rs:33:51
   |
33 | fn split_components(components: &[Component]) -> (&OsStr, &[Component]) {
   |                                                   ^ expected lifetime parameter
   |
   = help: this function's return type contains a borrowed value, but the signature does not say which one of `components`'s 2 lifetimes it is borrowed from
```

src/nodes/dir.rs
jmmv: The same would apply to the UID/GID, but I guess those are less of a bother for the testing case. (All of them should be passed as arguments but I went for "simplicity".) Anyway, plumbed `now` as an argument.

src/nodes/dir.rs
jmmv: Done. Gets rid of one level of indentation, but not sure this looks much better though.

src/nodes/dir.rs
jmmv: Yep, done.

src/nodes/dir.rs
jmmv: Good question. Nowhere at the moment. The Go variant doesn't do this either so I'd rather not change the behavior at this point to keep them consistent. Added a TODO.

The reason this doesn't seem to be a problem is two-fold: first, during readdir, we never use these stale entries because they are not marked as `explicit_mapping` so they never leak; and, second, even if we encounter any stale entry at lookup time we'll hit an `ENOENT` when doing the stat there, and lookup is a prerequisite for any other operation on the entry.

So... I don't know if this problem is even observable from the outside, but I may be missing some corner case. Would be nice to fix.

PR #41, opened 2018-10-24
Add handles for directories

jmmv:
The previous implementation used a single readdir hook to hande reads
from directories but this prevents supporting the case of reading from
an open-but-deleted directory -- which we need to support.

To fix this, add handle tracking to directories and implement separate
opendir/readdir operations for them.

There are integration tests that cover this functionality, but they are
coupled with the presence of setattr.  That's how I discovered that I
had to implement directory handles, but this means that I cannot enable
those tests until setattr lands (next PR).

### DIFF
--- a/src/nodes/dir.rs
+++ b/src/nodes/dir.rs
@@ -17,10 +17,12 @@ extern crate time;
 
 use {Cache, IdGenerator};
 use failure::{Error, ResultExt};
-use nix::{errno, unistd};
-use nodes::{KernelError, Node, NodeResult, conv};
+use nix::{errno, fcntl, sys, unistd};
+use nix::dir as rawdir;
+use nodes::{Handle, KernelError, Node, NodeResult, conv};
 use std::collections::HashMap;
 use std::ffi::{OsStr, OsString};
+use std::os::unix::ffi::OsStrExt;
 use std::fs;
 use std::io;
 use std::path::{Component, Path, PathBuf};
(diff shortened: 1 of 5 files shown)


jmmv:
For the most part, note that this is a code shuffle: the long implementation of `readdir` barely changes, except for using the API exposed by `nix`'s `Dir`; and the code in `lib.rs` is just moved to more-generic functions.

src/nodes/mod.rs
```diff
@@ -61,13 +62,41 @@ impl From<io::Error> for KernelError {
...
+    /// While this takes a `fuse::ReplyDirectory` object as a parameter for efficiency reasons, it
+    /// is the responsibility of the caller to invoke `reply.ok()` and `reply.error()` on the same
+    /// reply object.  This is for consistency with the handling of any errors returned by this and
+    /// other functions.
+    ///
+    /// `_ids` and `_cache` are the file system-wide bookkeeping objects needed to instantiate new
+    /// nodes, used when readdir discovers an underlying node that was not yet known.
+    fn readdir(&self, _ids: &IdGenerator, _cache: &Cache, _reply: &mut fuse::ReplyDirectory)
+        -> NodeResult<()> {
+        panic!("Not implemented");
```
taralx: Return ENOTDIR by default.

src/nodes/dir.rs
```diff
@@ -39,6 +41,99 @@ fn split_components<'a>(components: &'a [Component<'a>]) -> (&'a OsStr, &'a [Com
...
+    inode: u64,
+    writable: bool,
+    state: Arc<Mutex<MutableDir>>,
+
+    /// Handle for the open directory file descriptor.  This is `None` if the directory does not
+    /// have an underlying path.
+    handle: Mutex<Option<rawdir::Dir>>,
+}
+
+impl Handle for OpenDir {
```
taralx: Override other methods with EISDIR. It is very possible that the kernel does not type-check these things for you, so you need to guard against it.

src/nodes/mod.rs
jmmv: I see what you are trying to guard against, but I don't think the kernel should ever invoke this behavior. I've added an integration test instead.

src/nodes/dir.rs
jmmv: Same as above. Added a test.

src/nodes/mod.rs
jmmv: ... and even if it does, it seems that the FUSE libraries are guarding against it based on some inconsistencies I have observed in return error codes between macOS/Linux and Go/Rust. So I don't think I should have code that is never executed -- and if it _is_ ever executed, I'd like to know why.

jmmv:
PTAL?

PR #42, opened 2018-10-25
Implement setattr

jmmv:
This is the first mutable API implemented on nodes and unlocks support
for things like chmod, chown, chtimes, and truncate.


src/nodes/conv.rs
```diff
@@ -50,6 +51,12 @@ fn system_time_to_timespec(path: &Path, name: &str, time: &io::Result<SystemTime
     }
 }
 
+/// Converts a `time::Timespec` object into a `sys::time::TimeVal`.
+pub fn timespec_to_timeval(spec: Timespec) -> sys::time::TimeVal {
```
taralx: Ugh, this function makes me cringe. 😢 But I think there's no better way. It's annoying that TimeVal is (needlessly) opaque and doesn't have a sec+nsec constructor.

src/nodes/conv.rs
```diff
@@ -50,6 +51,12 @@ fn system_time_to_timespec(path: &Path, name: &str, time: &io::Result<SystemTime
...
+    const S_TO_NS: i64 = 1000 * 1000 * 1000;
+    sys::time::TimeVal::nanoseconds(spec.sec * S_TO_NS + i64::from(spec.nsec))
```
taralx: Why `::from`?

I recommend doing this by composition: `TimeVal::seconds(spec.sec) + TimeVal::nanoseconds(spec.nsec)`

src/nodes/dir.rs
```diff
@@ -247,6 +247,27 @@ impl Dir {
         }
         Dir::new_empty(ids.next(), Some(self), now)
     }
+
+    /// Same as `getattr` but with the node already locked.
+    fn getattr_unlocked(inode: u64, state: &mut MutableDir) -> NodeResult<fuse::FileAttr> {
+        let new_attr = match state.underlying_path.as_ref() {
```
taralx: Rust can partition borrows across fields:

```rust
        if let Some(path) = &state.underlying_path {
            let fs_attr = fs::symlink_metadata(path)?;
            if !fs_attr.is_dir() {
                warn!("Path {:?} backing a directory node is no longer a directory; got {:?}",
                    path, fs_attr.file_type());
                return Err(KernelError::from_errno(errno::Errno::EIO));
            }
            state.attr = conv::attr_fs_to_fuse(path, inode, &fs_attr);
        };
```

src/nodes/dir.rs
```diff
@@ -363,4 +367,12 @@ impl Node for Dir {
             handle: Mutex::from(handle),
         }))
     }
+
+    fn setattr(&self, delta: &AttrDelta) -> NodeResult<fuse::FileAttr> {
+        let mut state = self.state.lock().unwrap();
+        if let Some(path) = state.underlying_path.as_ref() {
```
taralx: Suggest using `&` here instead of `as_ref`.

src/nodes/mod.rs
```diff
@@ -74,9 +75,71 @@ impl From<nix::Error> for KernelError {
...
+    let mut result = Ok(());
+
+    if let Some(mode) = delta.mode {
+        result = result.and(
+            sys::stat::fchmodat(None, path, mode, sys::stat::FchmodatFlags::NoFollowSymlink));
+    }
+
+    if delta.uid.is_some() || delta.gid.is_some() {
+        result = result.and(
+            unistd::fchownat(None, path, delta.uid, delta.gid,
```
taralx: I don't suggest supporting chown/chgrp.

src/lib.rs
```diff
@@ -429,11 +431,38 @@ impl fuse::Filesystem for SandboxFS {
...
-        _bkuptime: Option<Timespec>, _flags: Option<u32>, _reply: fuse::ReplyAttr) {
-        panic!("Required RW operation not yet implemented");
+        _bkuptime: Option<Timespec>, _flags: Option<u32>, reply: fuse::ReplyAttr) {
+        let node = self.find_node(inode);
+        if !node.writable() {
+            reply.error(Errno::EACCES as i32);
+            return;
+        }
+
+        let nix_mode = mode.map(|m| sys::stat::Mode::from_bits_truncate(m as sys::stat::mode_t));
```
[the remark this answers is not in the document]
jmmv: Good idea. Done.

src/nodes/conv.rs
jmmv: I've added a TODO to get this addressed upstream.

(There are way too many time representations involved here :-/ )

src/nodes/conv.rs
jmmv: Nice. Done.

Before you ask, the `into()` is necessary because `spec.nsec` is an `i32` but the nanoseconds constructor expects an `i64`.

src/nodes/dir.rs
jmmv: Excellent. That was my expectation (that borrowing `underlying_path` shouldn't have an effect on `attr`)... but I can't quite understand why your simpler rewrite makes this work.

AFAICT the critical difference is between my previous `state.underlying_path.as_ref()` and your `&state.underlying_path`, right? Both borrows are on the field...

src/nodes/dir.rs
jmmv: Done.

src/nodes/mod.rs
jmmv: Heh, interesting thought. What's your rationale though?

(I'm going to leave this in for now given that the Go version supports it and there are integration tests covering this case. But I could add a TODO.)

src/nodes/dir.rs
taralx: I don't think that's the critical difference, but I can't see the old code to check any more.

src/nodes/mod.rs
taralx: It's just an invitation to security bugs. Non-privileged workloads can almost never use chown (and rarely use chgrp) legitimately anyway.

src/nodes/dir.rs
```diff
@@ -247,6 +247,21 @@ impl Dir {
         }
         Dir::new_empty(ids.next(), Some(self), now)
     }
+
+    /// Same as `getattr` but with the node already locked.
+    fn getattr_unlocked(inode: u64, state: &mut MutableDir) -> NodeResult<fuse::FileAttr> {
```
taralx: These kinds of functions are usually suffixed with `_locked`, aren't they?

src/nodes/dir.rs
jmmv: The old code did this:

```rust
        let new_attr = match state.underlying_path.as_ref() {
            Some(path) => {
                let fs_attr = fs::symlink_metadata(path)?;
                if !fs_attr.is_dir() {
                    warn!("Path {:?} backing a directory node is no longer a directory; got {:?}",
                        path, fs_attr.file_type());
                    return Err(KernelError::from_errno(errno::Errno::EIO));
                }
                Some(conv::attr_fs_to_fuse(path, inode, &fs_attr))
            },
            None => None,
        };
        if let Some(new_attr) = new_attr {
            state.attr = new_attr;
        }
```

(This time around note that I pushed review changes as new commits instead of amending the previous ones, precisely to avoid things vanishing during review time. I could revive these in the UI by clicking on the diff for the first commit. I've been told that's how GitHub PRs should be used... which is annoying because it doesn't work for commit chains.)

src/nodes/mod.rs
jmmv: Now that I think of it, the reason this can't be removed is because I'm planning on (also) using sandboxfs to contain the build of pkgsrc packages (via https://github.com/jmmv/pkg_comp, which currently uses bindfs). Building arbitrary software requires root privileges, and the builds actually end up calling into chown/chgrp...

src/nodes/dir.rs
jmmv: Yes. Dunno what I was thinking. Fixed the name.

src/nodes/mod.rs
taralx: Debian uses fakeroot to avoid needing privileges...

PR #44, opened 2018-10-31
Implement creation of all supported file types

jmmv:
This bag of commits adds support for creating all supported file types within directories.

If the PR is too long I can break it into pieces, but the reason I sent it in one chunk is because most commits are "trivial" once the refactorings that "Implement file creation and file writes" does are put in place.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -389,9 +389,21 @@ impl fuse::Filesystem for SandboxFS {
         }
     }
 
-    fn mkdir(&mut self, _req: &fuse::Request, _parent: u64, _name: &OsStr, _mode: u32,
-        _reply: fuse::ReplyEntry) {
-        panic!("Required RW operation not yet implemented");
+    fn mkdir(&mut self, _req: &fuse::Request, parent: u64, name: &OsStr, mode: u32,
+        reply: fuse::ReplyEntry) {
+        let dir_node = self.find_node(parent);
+        if !dir_node.writable() {
+            reply.error(Errno::EPERM as i32);
+            return;
+        }
+
+        match dir_node.mkdir(name, mode, &self.ids, &self.cache) {
+            Ok((node, attr)) => {
+                self.insert_node(node);
+                reply.entry(&TTL, &attr, IdGenerator::GENERATION);
+            },
+            Err(e) => reply.error(e.errno_as_i32()),
+        }
     }
 
     fn mknod(&mut self, _req: &fuse::Request, _parent: u64, _name: &OsStr, _mode: u32, _rdev: u32,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -406,9 +406,21 @@ impl fuse::Filesystem for SandboxFS {
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -291,6 +298,23 @@ impl SandboxFS {
...
+        let mut handles = self.handles.lock().unwrap();
+        debug_assert!(!handles.contains_key(&fh));
+        handles.insert(fh, handle);
+        fh
+    }
+
+    /// Tracks a node, which may already be known.
+    fn insert_node(&mut self, node: Arc<nodes::Node>) {
+        let mut nodes = self.nodes.lock().unwrap();
+        if !nodes.contains_key(&node.inode()) {
```
taralx: `nodes.entry(node.inode()).or_insert(node)`

taralx: (This effectively can make a one-liner if you want it to be...)

src/nodes/conv.rs
```diff
@@ -154,18 +155,47 @@ pub fn attr_fs_to_fuse(path: &Path, inode: u64, attr: &fs::Metadata) -> fuse::Fi
...
 
+/// Converts a set of `flags` bitmask to an `fs::OpenOptions`.
+///
+/// `allow_writes` indicates whether the file to be opened supports writes or not.  If the flags
+/// don't match this condition, then this returns an error.
+pub fn flags_to_openoptions(flags: u32, allow_writes: bool) -> NodeResult<fs::OpenOptions> {
+    let flags = flags as i32;
+    let mut options = fs::OpenOptions::new();
+    options.read(true);
+    if flags & (fcntl::OFlag::O_WRONLY | fcntl::OFlag::O_RDWR).bits() != 0 {
```
taralx: I feel like this would read better if you `fcntl::OFlag::from_bits_truncate` and then test `.contains(...)`.

src/nodes/dir.rs
```diff
@@ -262,6 +263,91 @@ impl Dir {
...
+    // Same as `lookup` but with the node already locked.
+    fn lookup_locked(writable: bool, state: &mut MutableDir, name: &OsStr, ids: &IdGenerator,
+        cache: &Cache) -> NodeResult<(Arc<Node>, fuse::FileAttr)> {
+        if let Some(dirent) = state.children.get(name) {
+            let refreshed_attr = dirent.node.getattr()?;
+            return Ok((dirent.node.clone(), refreshed_attr))
+        }
+
+        let (child, attr) = {
+            let path = match state.underlying_path.as_ref() {
```
taralx: Avoid using as_ref in this pattern -- it obscures what's going on.

src/lib.rs
jmmv: Done. Kept the function for consistency though.

src/nodes/conv.rs
jmmv: Done.

src/nodes/dir.rs
jmmv: Done.

PR #45, opened 2018-11-02
Implement node deletions

jmmv:
This PR is a complex because of all the required refactoring of `setattr`, but the implementation of the actual unlinking operations is pretty simple once that is done. Let me know if you want me to send the two commits separately for review.

Also, I've left a couple of TODOs in place because I just can't figure out what the type of the lambda should be. Clippy implies that I can remove the redundant closure, but that's not true.

And... we are almost feature complete! The most significant thing left is the ability to apply reconfigurations.

Thanks

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,6 +12,13 @@
 // License for the specific language governing permissions and limitations
 // under the License.
 
+// For portability reasons, we need to be able to cast integer values to system-level opaque
+// types such as "mode_t".  Because we don't know the size of those integers on the platform we
+// are building for, sometimes the casts do widen the values but other times they are no-ops.
+#![cfg_attr(feature = "cargo-clippy", allow(identity_conversion))]
+
+// We construct complex structures in multiple places, and allowing for redundant field names
+// increases readability.
 #![cfg_attr(feature = "cargo-clippy", allow(redundant_field_names))]
 
 #[macro_use] extern crate failure;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -489,9 +489,17 @@ impl fuse::Filesystem for SandboxFS {
         panic!("Required RW operation not yet implemented");
     }
 
-    fn rmdir(&mut self, _req: &fuse::Request, _parent: u64, _name: &OsStr,
-        _reply: fuse::ReplyEmpty) {
-        panic!("Required RW operation not yet implemented");
+    fn rmdir(&mut self, _req: &fuse::Request, parent: u64, name: &OsStr, reply: fuse::ReplyEmpty) {
+        let dir_node = self.find_node(parent);
+        if !dir_node.writable() {
+            reply.error(Errno::EPERM as i32);
+            return;
+        }
+
+        match dir_node.rmdir(name) {
+            Ok(_) => reply.ok(),
+            Err(e) => reply.error(e.errno_as_i32()),
+        }
     }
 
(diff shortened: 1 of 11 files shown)


src/nodes/mod.rs
```diff
@@ -88,57 +88,135 @@ pub struct AttrDelta {
...
-            // TODO(jmmv): Should use futimensat to avoid following symlinks but this function is
-            // not available on macOS El Capitan, which we currently require for CI builds.
-            warn!("Asked to update atime/mtime for symlink {} but following symlink instead",
-                path.display());
-        }
-        result = result.and(sys::stat::utimes(path, &atime, &mtime));
+    let result = try_path(path, |p|
+        unistd::fchownat(None, p, uid, gid, unistd::FchownatFlags::NoFollowSymlink));
+    if result.is_ok() {
+        attr.uid = uid.map_or_else(|| attr.uid, u32::from);
```
taralx: Does `map_or` work here? attr.uid is probably Copy, so it shouldn't cause an issue.

src/nodes/mod.rs
```diff
@@ -88,57 +88,135 @@ pub struct AttrDelta {
...
+    atime: Option<sys::time::TimeVal>, mtime: Option<sys::time::TimeVal>)
+    -> Result<(), nix::Error> {
+    if atime.is_none() && mtime.is_none() {
+        return Ok(());
+    }
+
+    let atime = atime.unwrap_or_else(|| conv::timespec_to_timeval(attr.atime));
+    let mtime = mtime.unwrap_or_else(|| conv::timespec_to_timeval(attr.mtime));
+    if attr.kind == fuse::FileType::Symlink {
+        // TODO(jmmv): Should use futimensat to avoid following symlinks but this function is
```
taralx: Not a fan of this. We should at some point conditionalize and this path should fail if you try to do this to a symlink.

taralx:
No new push?

jmmv:
I think I got distracted with necessary `nix` changes to make this work well and forgot about getting back to the review. Pushed now but can't remember what status this was in; hopefully nothing embarrassing.

FTR, this PR requires https://github.com/nix-rust/nix/pull/968 in order to get the Linux builds working. And https://github.com/nix-rust/nix/pull/967 will hopefully let me address the issue of updating the times of symlinks.

jmmv:
Alright, https://github.com/nix-rust/nix/pull/968 was merged into nix so I have updated `Cargo.toml` accordingly here and fixed the builds. They are now green so this should be ready for another pass. Thanks!

src/lib.rs
```diff
@@ -12,6 +12,13 @@
...
 // under the License.
 
+// For portability reasons, we need to be able to cast integer values to system-level opaque
+// types such as "mode_t".  Because we don't know the size of those integers on the platform we
+// are building for, sometimes the casts do widen the values but other times they are no-ops.
+#![cfg_attr(feature = "cargo-clippy", allow(identity_conversion))]
+
+// We construct complex structures in multiple places, and allowing for redundant field names
+// increases readability.
 #![cfg_attr(feature = "cargo-clippy", allow(redundant_field_names))]
```
taralx: Filed https://github.com/rust-lang-nursery/rust-clippy/issues/3422 for this.

jmmv: Thanks! Seems like a good idea.

PR #49, opened 2018-11-20
Make the TTL of file metadata configurable

jmmv:
Add a new --ttl flag to let the user of sandboxfs configure how long
to cache file metadata for in the kernel before the kernel decides to
query the user-space daemon again.

This feature will come in handy when implementing reconfigurations
because we won't be able to implement unmaps with correct semantics
just yet: the FUSE bindings we use do not support kernel invalidations
so we will tell the user to adjust the TTL instead.  (I want to add
the missing feature to the FUSE bindings... but it's not a trivial
effort.)

Note that this feature is not a hack per se: exposing the TTL value
to the user as a configuration setting is worthwhile and so this is
here to stay.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,6 +16,7 @@ extern crate env_logger;
 #[macro_use] extern crate failure;
 extern crate getopts;
 extern crate sandboxfs;
+extern crate time;
 
 use failure::Error;
 use getopts::Options;
@@ -23,12 +24,43 @@ use std::env;
 use std::path::{Path, PathBuf};
 use std::process;
 use std::result::Result;
+use time::Timespec;
+
+/// Default value of the `--ttl` flag.
+///
+/// This is expressed as a string rather than a parsed value to ensure the default value can be
+/// parsed with the same semantics as user-provided values.
+static DEFAULT_TTL: &str = "60s";
+
+/// Suffix for durations expressed in seconds.
+static SECONDS_SUFFIX: &str = "s";
 
 /// Execution failure due to a user-triggered error.
 #[derive(Debug, Fail)]
 #[fail(display = "{}", message)]
 struct UsageError {
     message: String,
+}
+
+/// Parses the value of a flag that takes a duration.
+///
+/// At the moment, the only supported unit for the duration are seconds and the textual value
+/// must include the "ms" suffix to support adding extra units in the future if desired.
+fn parse_duration(value: &str) -> Result<Timespec, UsageError> {
+    if !value.ends_with(SECONDS_SUFFIX) {
+        let message = format!("invalid time specification {}: must end in 's'", value);
+        return Err(UsageError { message });
+    }
+    let value = &value[0..value.len() - SECONDS_SUFFIX.len()];
+    match value.parse::<u32>() {
+        Ok(sec) => {
+            Ok(Timespec { sec: i64::from(sec), nsec: 0 })
+        },
+        Err(e) => {
+            let message = format!("invalid time specification {}: {}", value, e);
+            Err(UsageError { message })
+        }
+    }
 }
 
 /// Takes the list of strings that represent mappings (supplied via multiple instances of the
@@ -107,6 +139,9 @@ fn safe_main(program: &str, args: &[String]) -> Result<(), Error> {
     let mut opts = Options::new();
     opts.optflag("", "help", "prints usage information and exits");
     opts.optmulti("", "mapping", "type and locations of a mapping", "TYPE:PATH:UNDERLYING_PATH");
+    opts.optopt("", "ttl",
+        &format!("how long the kernel is allowed to keep file metadata (default: {})", DEFAULT_TTL),
+        &format!("TIME{}", SECONDS_SUFFIX));
     let matches = opts.parse(args)?;
 
     if matches.opt_present("help") {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -31,6 +42,27 @@ struct UsageError {
     message: String,
 }
 
+/// Parses the value of a flag that takes a duration.
+///
+/// At the moment, the only supported unit for the duration are seconds and the textual value
+/// must include the "ms" suffix to support adding extra units in the future if desired.
+fn parse_duration(value: &str) -> Result<Timespec, UsageError> {
+    if !value.ends_with(SECONDS_SUFFIX) {
+        let message = format!("invalid time specification {}: must end in 's'", value);
```
kevincox: If you are going to put `SECONDS_SUFFIX` in a constant I would also use the constant in the error string so that they stay in sync.

src/main.rs
```diff
@@ -31,6 +42,27 @@ struct UsageError {
...
+        return Err(UsageError { message });
+    }
+    let value = &value[0..value.len() - SECONDS_SUFFIX.len()];
```
kevincox: I would add parens because I feel that having no space after the `..` makes it bind more tightly in my brain so I prefer to include them. You can also omit the lower bound.
```suggestion
    let value = &value[..(value.len() - SECONDS_SUFFIX.len())];
```

src/main.rs
```diff
@@ -31,6 +42,27 @@ struct UsageError {
...
+    match value.parse::<u32>() {
+        Ok(sec) => {
+            Ok(Timespec { sec: i64::from(sec), nsec: 0 })
+        },
+        Err(e) => {
+            let message = format!("invalid time specification {}: {}", value, e);
+            Err(UsageError { message })
+        }
+    }
```
kevincox: Personally I have a slight preference for the following. However what you have is also readable.

```rust
value.parse::<u32>()
  .map(|sec| Timespec { sec: i64::from(sec), nsec: 0 })
  .map_err(|e| UsageError { message: format!("invalid time specification {}: {}", value, e) })
```

src/main.rs
```diff
@@ -116,13 +151,19 @@ fn safe_main(program: &str, args: &[String]) -> Result<(), Error> {
...
+        Some(value) => parse_duration(&value)?,
+        None => parse_duration(DEFAULT_TTL).expect(
+            "default value for flag is not accepted by the parser; this is a bug in the value"),
+    };
+
     let mount_point = if matches.free.len() == 1 {
         &matches.free[0]
     } else {
         return Err(UsageError { message: "invalid number of arguments".to_string() }.into());
     };
```
kevincox: I realize this is existing code but I would recommend:

```rust
let mount_point = if let Some(mount_point) = matches.free.first() {
  mount_point
} else {
  return Err(UsageError { message: "invalid number of arguments".to_string() }.into());
};
```

src/main.rs
```diff
@@ -165,6 +206,32 @@ mod tests {
     use sandboxfs::Mapping;
     use super::*;
 
+    /// Checks that an error, once formatted for printing, contains the given substring.
+    fn err_contains<F: failure::Fail>(substr: &str, err: F) {
```
kevincox:
```suggestion
    fn err_contains(substr: &str, err: impl failure::Fail) {
```

src/main.rs
```diff
@@ -165,6 +206,32 @@ mod tests {
...
+    #[test]
+    fn test_parse_duration_ok() {
+        assert_eq!(Timespec { sec: 1234, nsec: 0 }, parse_duration("1234s").unwrap());
+    }
+
+    #[test]
+    fn test_parse_duration_bad_unit() {
+        err_contains("must end in 's'", parse_duration("1234").unwrap_err());
+        err_contains("invalid digit", parse_duration("1234ms").unwrap_err());
+        err_contains("invalid digit", parse_duration("1234ss").unwrap_err());
```
kevincox: This isn't really the best error.

src/main.rs
```diff
@@ -31,6 +42,27 @@ struct UsageError {
     message: String,
 }
 
+/// Parses the value of a flag that takes a duration.
+///
+/// At the moment, the only supported unit for the duration are seconds and the textual value
+/// must include the "ms" suffix to support adding extra units in the future if desired.
+fn parse_duration(value: &str) -> Result<Timespec, UsageError> {
```
kevincox: This implementation of parser is harder to extend to new units and makes it difficult to give good error messages. Why not do something like the following to define the protocol between the value and units so that you can say things like "invalid unit 'ms'"?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=34710115ee2ea47323746c1bf8d1f011

src/main.rs
jmmv: Heh, of course. The constant was a later addition and apparently I missed this case. Irrelevant after all other changes though.

src/main.rs
jmmv: Not applicable any longer.

src/main.rs
jmmv: Nice. Wasn't familiar with this idiom and I like that it's more concise.

src/main.rs
jmmv: Done.

src/main.rs
jmmv: Fixed this instance. But following your new proposal, some other errors have become not great either. Don't care much though.

src/main.rs
```diff
@@ -31,6 +42,26 @@ struct UsageError {
...
+    let (value, unit) = match s.find(|c| !char::is_ascii_digit(&c) && c != '-') {
+        Some(pos) => s.split_at(pos),
+        None => {
+            let message = format!("invalid time specification {}: missing unit", s);
+            return Err(UsageError { message });
+        },
+    };
+
+    if unit != SECONDS_SUFFIX {
+        let message = format!("invalid time specification {}: unsupported unit '{}'", s, unit);
```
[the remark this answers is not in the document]
jmmv: Good point. Done.

PR #50, opened 2018-11-28
Implement termination signal handling

jmmv:
Upon receiving a termination signal, sandboxfs should unmount the file
system instead of abruptly exiting and leaving the mount point behind
in an unusable state.  Do so now.

At first I cloned the exact implementation of the Go variant, using the
same message-passing structure between the main program and the thread
handling the signals.  This worked fine but then realized (thanks to a
"note" in the original code!) that we can do better in Rust because of
the ability to control the signal mask via the nix crate.

Note, however, that the way signals are handled may still seem odd in
cases (like e.g. ignoring all signals after the first one is received).
This is intentional to keep the behavior in sync with the Go code and
with what is currently tested.

### DIFF
--- a/src/concurrent.rs
+++ b/src/concurrent.rs
@@ -12,12 +12,20 @@
 // License for the specific language governing permissions and limitations
 // under the License.
 
+use failure::Error;
 use nix::unistd;
+use nix::sys::signal;
+use signal_hook;
+use std::cmp;
 use std::fs;
-use std::io;
+use std::io::{self, Read};
 use std::os::unix::io as unix_io;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::mpsc;
+use std::time;
+use std::thread;
 
 /// A file with a single scoped owner but with multiple non-owner views.
 ///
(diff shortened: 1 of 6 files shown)


src/concurrent.rs
```diff
@@ -103,6 +113,145 @@ impl io::Read for ShareableFile {
...
+/// doesn't matter: we have entered a terminal status: we do this at exit time so we'll keep trying
+/// to unclog things while telling the user what's going on.  They are the ones that have to fix
+/// this situation.
+fn retry_unmount<P: AsRef<Path>>(mount_point: P) {
+    let mut backoff = time::Duration::from_millis(10);
+    'retry: loop {
+        match fuse::unmount(mount_point.as_ref()) {
+            Ok(()) => break 'retry,
+            Err(e) => {
+                warn!("Unmounting file system failed with error '{}'; will retry in {:?}",
```
taralx: Don't warn until backoff is 1s. Otherwise you get a spam of these initially before the user can really do anything about it.

src/concurrent.rs
```diff
@@ -103,6 +113,145 @@ impl io::Read for ShareableFile {
...
+                    e, backoff);
+                thread::sleep(backoff);
+                if backoff.as_secs() < 1 {
```
taralx: If you want to cap at 1s and not 1.28s, you need a min function here.

src/concurrent.rs
jmmv: Done.

src/concurrent.rs
jmmv: Done.

PR #51, opened 2018-12-04
Make ShareableFile actually unblock reads

jmmv:
As it turns out, closing the file descriptor on which a read syscall is
blocked doesn't actually cause the read to unblock.  According to some
documentation I found online for Linux, this is because the read blocks
on a the file itself so closing the descriptor has no effect on the read.
I suspect a similar reason applies to macOS, where I'm seeing this now.

I have no clue how I concluded that this worked in the past.  Obviously
the test was very wrong (writing to the FIFO actually made the read
unblock regardless of what we did elsewhere)... but I had tested this on
the real sandboxfs code after implementing reconfigurations and things
seemed fine.  I was probably doing the wrong thing too, and now that
singal handling is in place, the problem becomes more obvious.

### DIFF
--- a/src/concurrent.rs
+++ b/src/concurrent.rs
@@ -14,7 +14,7 @@
 
 use failure::Error;
 use nix::unistd;
-use nix::sys::signal;
+use nix::sys::{self, signal};
 use signal_hook;
 use std::cmp;
 use std::fs;
@@ -27,12 +27,20 @@ use std::sync::mpsc;
 use std::time;
 use std::thread;
 
-/// A file with a single scoped owner but with multiple non-owner views.
+/// Converts a `nix::Error` that we expect to carry an errno to an `io::Error`.
+fn nix_to_io_error(err: nix::Error) -> io::Error {
+    match err {
+        nix::Error::Sys(errno) => io::Error::from_raw_os_error(errno as i32),
+        e => panic!("Did not expect to get an error without an errno from a nix call: {:?}", e),
+    }
+}
+
+/// A scope-owned file with support for multiple non-owner readers in different threads.
 ///
 /// A `ShareableFile` object owns the file passed to it at construction time and will close the
 /// underlying file handle when the object is dropped.
 ///
-/// Concurrent views into this same file, obtained via the `clone_unowned` method, do not own the
+/// Concurrent views into this same file, obtained via the `reader` method, do not own the
 /// file handle.  Such views must accept the fact that the handle can be closed at any time, a
 /// condition that is simply exposed as if the handle reached EOF.  Concurrent views can safely be
 /// moved across threads.
@@ -41,16 +49,15 @@ pub struct ShareableFile {
     /// Underlying file descriptor shared across all views of this file.
     fd: unix_io::RawFd,
 
-    /// Whether this instance of the file is responsible for closing the file descriptor.
-    owned: bool,
+    /// Write ends of the pipes used by `ShareableFileReader` to know if the file has been closed.
+    watchers: Vec<unix_io::RawFd>,
 
     /// Whether the file descriptor has already been closed or not.
     ///
     /// In principle, we don't need this field: concurrent readers of this file will get their
-    /// reads abruptly terminated with an error, and we should be able to rely on EBADFD to tell
-    /// that this happened because of us closing the file descriptor.  Unfortunately... macOS
-    /// Mojave doesn't cooperate and yields very strange error codes (like EISDIR) pretty frequently
-    /// (but not deterministically).  May it be an APFS bug?
+    /// selects abruptly terminated with an error and we should be able to rely on `EBADF` to tell
+    /// that this happened because of us closing the file descriptor.  But it seems better to track
+    /// our own closed condition so that we can distinguish actual errors from expected errors.
     closed: Arc<AtomicBool>,
 }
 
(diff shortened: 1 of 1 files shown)


src/concurrent.rs
```diff
@@ -289,33 +300,58 @@ mod tests {
...
-        // We need to write to the FIFO for the read-only open below to not block.
+        // We need to open the FIFO for writes so that the open for reads doesn't block.  But we
+        // have to do this open in a separate thread because it will block too.
+        let (writer_unblock_tx, writer_unblock_rx) = mpsc::channel();
         let writer_handle = {
             let path = path.clone();
             thread::spawn(move || {
-                let mut writer = fs::File::create(&path).unwrap();
-                write!(writer, "some text longer than we'll ever read")
+                let _keep_open = fs::File::create(path).unwrap();
```
kevincox: What I find more clear then relying on this living for the scope is to put an explicit [`std::mem::drop(keep_open)`](https://doc.rust-lang.org/std/mem/fn.drop.html) after the `recv` call.

src/concurrent.rs
```diff
@@ -289,33 +300,58 @@ mod tests {
...
+                reader_ready_tx.send(()).unwrap();
                 file.read(&mut buffer)
             })
         };
 
-        writer_handle.join().unwrap().expect("Write didn't finish successfully");
+        // This is racy: there is no guarantee that the reader thread is now blocked on the read
+        // syscall although there is a high likelihood that it is.  If we experience the race, the
+        // test will silently pass even if there is a problem in the ShareableFile.
+        reader_ready_rx.recv().unwrap();
```
kevincox: Does it make sense to add a small sleep to raise the probability of blocking before this line is triggered? I can imagine that if you are running the tests on a single core machine the scheduler could predictably prefer one thread over the other. Even a small sleep should make that unlikely to always occur on any machine.

src/concurrent.rs
```diff
@@ -289,33 +300,58 @@ mod tests {
...
 
-    #[test]
-    fn test_shareable_file_close_unblocks_reads_without_error() {
+    fn try_shareable_file_close_unblocks_reads_without_error() {
         let dir = tempfile::tempdir().unwrap();
         let path = dir.path().join("pipe");
         unistd::mkfifo(&path, sys::stat::Mode::S_IRUSR | sys::stat::Mode::S_IWUSR).unwrap();
 
-        // We need to write to the FIFO for the read-only open below to not block.
+        // We need to open the FIFO for writes so that the open for reads doesn't block.  But we
```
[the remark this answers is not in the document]
jmmv: Yeah, but then I don't think we would be testing what I need to test :)

We need the read to block because that's how the reconfiguration thread will behave, and the whole point of this test is verifying that we can unblock that read asynchronously. If I opened the file in non-blocking mode, that wouldn't be the case right?

src/concurrent.rs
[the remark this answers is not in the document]
jmmv: Hmm... I don't think I can open the writer on the main thread because it would block, but your comment gave me an idea -- or maybe that's what you were proposing after all. I can just return the result of create() here and handle any value in the parent thread.

jmmv: Interesting. Redid this a bit though to simplify what the thread does, but still applied your suggestion.

src/concurrent.rs
jmmv: Yes. Done.

This had crossed my mind but for some reason I was thinking "I have to sleep at least one second, and that'd be too annoying on unit tests". But 1ms is fine.

src/concurrent.rs
```diff
@@ -289,33 +300,58 @@ mod tests {
...
                 file.read(&mut buffer)
             })
         };
 
-        writer_handle.join().unwrap().expect("Write didn't finish successfully");
+        // This is racy: there is no guarantee that the reader thread is now blocked on the read
+        // syscall although there is a high likelihood that it is.  If we experience the race, the
+        // test will silently pass even if there is a problem in the ShareableFile.
+        reader_ready_rx.recv().unwrap();
+
```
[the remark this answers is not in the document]
jmmv: Glad you reviewers agree on the same points :) Done.

src/concurrent.rs
taralx: You want the sleep after the receive.

jmmv: Oops, done.

jmmv:
OOOOOOOMG. What a nightmare this is turning into.

Turns out this was still insufficient for Linux -- but thanks Travis for catching the problem. Fortunately, the Linux `select(2)` manpage is more helpful in this case:

```
   Multithreaded applications
       If a file descriptor being monitored by select() is closed  in  another
       thread,  the  result  is  unspecified.   On some UNIX systems, select()
       unblocks and returns, with an indication that the  file  descriptor  is
       ready  (a  subsequent  I/O  operation  will  likely fail with an error,
       unless another the file descriptor reopened between the  time  select()
       returned  and  the  I/O  operations was performed).  On Linux (and some
       other systems), closing the file descriptor in another  thread  has  no
       effect  on select().  In summary, any application that relies on a par-
       ticular behavior in this scenario must be considered buggy.
```

So... had to add an extra pipe so that the main thread can tell readers of the close by writing into the pipe, not closing it from under the select call. This seems to work fine now on Linux and on macOS based on local testing. Hopefully Travis agrees.

taralx:
Right. Unless you want signals. Ah, well.

PR #54, opened 2018-12-11
Implement renames and moves

jmmv:
Oops. I completely forgot about implementing this operation. I am not too happy with the current code (especially because of handling moves and renames separately), but I wasn't either with the Go variant due to its introspection of types. Trying a different approach.

----- Commit message -----

Handling renames within a single directory is easy, but supporting a move
across directories is difficult because we need to lock two nodes and it
is good to not leak the implications of this through the Node trait.

To support the latter case, I have split the rename operation in two: one
for the responsibilities of the source node and one for the target node.
This approach allows keeping the Node interface and the Dir type
oblivious to the various Node types that exist (i.e. we don't need
introspection as the Go variant of this code needed).  The downside is
that the need for the source/target method split looks strange, and in
the source side, we have to support "rolling back" a failed move.
Having seen the two possible implementations, I think this is clearer.

### DIFF
--- a/src/nodes/dir.rs
+++ b/src/nodes/dir.rs
@@ -136,7 +136,8 @@ impl Handle for OpenDir {
 }
 
 /// Representation of a directory entry.
-struct Dirent {
+#[derive(Clone)]
+pub struct Dirent {
     node: ArcNode,
     explicit_mapping: bool,
 }
@@ -149,7 +150,7 @@ pub struct Dir {
 }
 
 /// Holds the mutable data of a directory node.
-struct MutableDir {
+pub struct MutableDir {
     parent: u64,
     underlying_path: Option<PathBuf>,
     attr: fuse::FileAttr,
@@ -387,6 +388,13 @@ impl Node for Dir {
         state.underlying_path = None;
     }
 
+    fn set_underlying_path(&self, path: &Path) {
+        let mut state = self.state.lock().unwrap();
+        debug_assert!(state.underlying_path.is_some(),
+            "Renames should not have been allowed in scaffold or deleted nodes");
+        state.underlying_path = Some(PathBuf::from(path));
+    }
+
     fn map(&self, components: &[Component], underlying_path: &Path, writable: bool,
         ids: &IdGenerator, cache: &Cache) -> Result<(), Error> {
         let (name, remainder) = split_components(components);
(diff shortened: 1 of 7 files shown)


src/nodes/dir.rs
```diff
@@ -518,6 +526,63 @@ impl Node for Dir {
...
+            // "Roll back" any changes we did to the current directory because the rename could not
+            // be completed on the target.
+            state.children.insert(old_name, dirent);
+        }
+        result
+    }
+
+    fn rename_and_move_target(&self, dirent: &Dirent, old_path: &Path, new_name: &OsStr)
+        -> NodeResult<()> {
+        // We are locking the target node while the source node is already locked, so there might be
```
taralx: This will deadlock on any attempt to handle operations in parallel. You do need to mimic Go here, with a lock ordering constraint. Inode number is fine, as is any other totally ordered relation that is bijective with inodes.

Thankfully, visibility lets you reach into the other directory entry, and you should. Have a single `rename_and_move` function that takes the locks and then does the work.

taralx: I believe `std::any::Any` will work here to allow you to downcast the Node.

jmmv: I had tried the downcast approach previously and couldn't get it to work. I did something like:

```
let new_dir = new_node.downcast::<Dir>();
```

but this complains about `downcast` not existing. I've seen somewhere that I may need to add an `as_any()` function to the `Node` to convert it to an `Any`, and then this could work here. Is that what has to happen?

(Eventually that's why I went with the current approach, to avoid the downcasts-that-didn't-work and having to expose `MutableDir` in the `Node` trait.)

By the way, I've added a new test (pushed here) to try to catch this potential deadlock... and haven't been able to. Even adding random sleeps after acquiring each lock is unable to catch the problem. I agree with you that the problem can exist though...

taralx: The magic you are looking for is `trait Node : Any`.

taralx: Is it possible there is some serialization happening at a higher level?

jmmv: Oh, duh... now that you mention it, the Rust FUSE library handles all requests sequentially... (Mind you, that's the primary reason it's way faster than the Go version.)

taralx: Oh, my. That changes everything. You don't need `Arc` anywhere, many things can just be borrows, etc. (More particularly, your `ArcNode` can just be `&mut dyn Node`.

jmmv: Well, I think I still do need `Arc`s and stuff because of the reconfigurations, which happen from a separate thread.

I'll still try to get the `Any` solution working and the lock ordering fixed before merging this PR though.

jmmv: Fair enough. I tried the `Any` thingy and couldn't get it to work so going with this for now. Updated the TODO that was in the code to reflect reality and to explain that the deadlock would indeed happen.

PR #56, opened 2018-12-18
Upgrade dependencies

jmmv:
The various changes in this PR upgrade all components of the `WORKSPACE` to newer versions and also make our Travis CI builds use a modern Bazel. These changes are all required to allow sandboxfs to use recent Bazel versions, as the old Go rules we used stopped working.

### DIFF
--- a/admin/lint/checks.go
+++ b/admin/lint/checks.go
@@ -31,7 +31,7 @@ import (
 
 // checkLicense checks if the given file contains the necessary license information and returns an
 // error if this is not true or if the check cannot be performed.
-func checkLicense(file string) error {
+func checkLicense(workspaceDir string, file string) error {
 	for _, pattern := range []string{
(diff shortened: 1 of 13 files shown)


admin/lint/checks.go
```diff
@@ -31,7 +31,7 @@ import (
 
 // checkLicense checks if the given file contains the necessary license information and returns an
 // error if this is not true or if the check cannot be performed.
-func checkLicense(file string) error {
+func checkLicense(workspaceDir string, file string) error {
```
jin: unrelated changes?

jmmv: No, these were necessary. The workspace directory changes were already in a separate commit with a justification for them (to fix Gazelle runs) and I have now also split out the test changes to explain why they are necessary (to pass with Go 1.11).

PR #61, opened 2019-01-02
Bye bye Go and Bazel

jmmv:
The primary goal behind these changes it the removal of the Go implementation of sandboxfs. We can do this now as the Rust variant of sandboxfs is now sufficiently functional to serve Bazel, our primary and only user.

But a large side-effect of these changes is the removal of our reliance on Bazel as well. I originally introduced this to "sanitize" the use of Go by avoiding the `GOPATH` and unversioned dependencies... but it has turned out to be a bad idea. I have spent a ton of time fighting Bazel for not a lot of gain given the tiny size of this project -- and now that we don't really need Go any longer, the issues I tried to resolve are less relevant. As a result of this, VSCode now works again, Git's pre-commit hook speed is tolerable again, and Travis CI builds are, at least, twice as fast.

Note that, even though this introduces a `Makefile` to orchestrate various aspects of the build, this project is primarily built by Cargo and I do not intend to change that. The `Makefile` is just an extra for the convenience of integrating the Go pieces with the rest of the project and should remain as such.

And also note that Go code remains. All integration tests are in Go and I have no plans to rewrite them.


man/sandboxfs.1
```diff
@@ -133,21 +106,18 @@ is not repeated.
...
 .It Fl -output Ar path
 Points to the file to which to write confirmations of reconfiguration, or
 .Sq -
 (the default) for stdout.
 See the
 .Sx Reconfigurations
 subsection for details on the contents and behavior of the output file.
+.It Fl -ttl Ar duration
+Specifies how long the kernel is allowed to cache file metadata for.
+The duration is specified as a number of seconds followed by the
```
taralx: Does it support any other units?

configure
```diff
@@ -100,6 +102,32 @@ setup_git() {
...
+  fi
+
+  if [ -z "${GOROOT}" ]; then
+    info "Go not found; cannot run integration tests"
+  else
+    info "Using Go with GOROOT: ${GOROOT}"
+
+    mkdir -p .gopath/src/github.com/bazelbuild
+    rm -f .gopath/src/github.com/bazelbuild/sandboxfs
+    ln -s ../../../.. .gopath/src/github.com/bazelbuild/sandboxfs
```
taralx: Not a fan of this. Deep relative paths are easy to get wrong and hard to review for correctness.

configure
```diff
@@ -100,6 +102,32 @@ setup_git() {
...
+
+    for dep in bazil.org/fuse golang.org/x/net/context golang.org/x/sys/unix; do
```
taralx: Strongly recommend moving to Go modules soon. :-)

admin/lint/checks.go
```diff
@@ -122,7 +122,7 @@ func checkGolint(workspaceDir string, file string) error {
 // checkAll runs all possible checks on a file.  Returns true if all checks pass, and false
 // otherwise.  Error details are dumped to stderr.
 func checkAll(workspaceDir string, file string) bool {
-	isBuildFile := filepath.Base(file) == "BUILD.bazel" || filepath.Ext(file) == ".bzl" || filepath.Base(file) == "Makefile.in"
+	isBuildFile := filepath.Base(file) == "Makefile.in"
```
taralx: Just inline this below now. It's only used to skip tab checks, which is specific to Makefiles and not build files generally.

man/sandboxfs.1
jmmv: Not at the moment. Added "currently" to clarify.

admin/lint/checks.go
jmmv: I was about to do this... but realized that it's also used as part of the definition of `isDocumentation`. I think it makes sense to keep it.

configure
jmmv: Would you use an absolute path instead? (Yes, this has bitten me many times, but somehow I like the idea of making the tree "self-contained" in case it's moved. Not that it'd work though... I'm sure there would be many more things that would break in a move.)

configure
jmmv: I'll have to read about that then. Haven't followed new developments much!

jmmv:
Yeah, not a fan of make either... but Bazel has proven to be a time sink so far and pretty intrusive; the addition of protobuf as a Gazelle dependency was the last straw.

If the Rust rules were up to the task I might have tried using them, but Cargo is so good it's hard to discard. And all this make stuff is for support tasks anyway so I'm not too worried about it being suboptimal: I plan to publish this to crates.io anyway and support the standard `cargo install`.

configure
taralx: I would probably try to use a variable-rooted path. Then the problem condenses down to being sure that the variable is correct, which is often easier if the variable is constructed in a structured fashion.

jmmv: Tried doing this... but a for loop to construct the directory and symlink target seemed more complex and less readable. After all, it's a single symlink so I think I'll keep this as is for now.

PR #63, opened 2019-01-05
Test sandboxfs with a real Bazel build

jmmv:
Add a Travis CI configuration to use Bazel with sandboxfs to build a reasonably-sized project (Bazel itself), because Bazel is our primary and only user for now. This will ensure sandboxfs is usable for at least some builds and that its interface remains compatible with Bazel.

This increases presubmit times from about 10 minutes to over 25. I'm sure this will become annoying but let's see how far we can keep it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,21 @@ env:
   - DO=lint
   - DO=test
 
+jobs:
+  include:
+    # Validating sandboxfs against a Bazel build is expensive (think over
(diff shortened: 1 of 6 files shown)


jmmv:
I've been playing with Travis cron jobs and found a way to configure this expensive test to only run on a daily basis. I think that's good enough and keeps iterative builds for PRs and pushes lean.

PR #66, opened 2019-01-15
Add CPU profiling support

jmmv:
These changes add support for generating CPU profiles that can later be inspected with gperftools. The actual code needed to make this happen is tiny, but making the dependency on gperftools conditional has been the hard thing (and I really want this because gperftools has to be installed by hand upfront and is a C library, which would make sandboxfs compilation even harder).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,10 @@ version = "0.1.0"
 [badges]
 travis-ci = { repository = "bazelbuild/sandboxfs", branch = "master" }
 
+[features]
+default = []
+profiling = ["cpuprofiler"]
+
 [dependencies]
 env_logger = "0.5"
 failure = "~0.1.2"
@@ -23,6 +27,13 @@ serde_derive = "1.0"
 serde_json = "1.0"
 signal-hook = "0.1"
 time = "0.1"
+
+[dependencies.cpuprofiler]
+# TODO(https://github.com/AtheMathmo/cpuprofiler/pull/10): Replace this with
+# 0.0.4 or an upstream branch reference once released.
+git = "https://github.com/jmmv/cpuprofiler.git"
+rev = "a852024d6aed7202863dc43a3348e793aa432d54"
+optional = true
 
 [dependencies.fuse]
 # TODO(jmmv): Replace this with 0.4 once released.
--- a/integration/utils/config.go
+++ b/integration/utils/config.go
@@ -17,10 +17,15 @@ package utils
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 )
 
 // Config represents the configuration for the integration tests as provided in the command line.
 type Config struct {
+	// Features contains the set of features enabled during the build.  This represents a set
+	// and therefore the values in the map are meaningless.
+	Features map[string]bool
+
 	// ReleaseBinary is true if the binary provided in SandboxfsBinary was built in release
 	// mode, false otherwise.
 	ReleaseBinary bool
@@ -41,9 +46,14 @@ var globalConfig *Config
 
 // SetConfigFromFlags initializes the test configuration based on the raw values provided by the
 // user on the command line.  Returns an error if any of those values is incorrect.
-func SetConfigFromFlags(releaseBinary bool, rawSandboxfsBinary string, unprivilegedUserName string) error {
+func SetConfigFromFlags(rawFeatures string, releaseBinary bool, rawSandboxfsBinary string, unprivilegedUserName string) error {
 	if globalConfig != nil {
 		panic("SetConfigFromFlags can only be called once")
+	}
+
+	features := make(map[string]bool)
+	for _, feature := range strings.Split(rawFeatures, " ") {
+		features[feature] = true
 	}
 
 	sandboxfsBinary, err := filepath.Abs(rawSandboxfsBinary)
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -24,6 +28,14 @@ serde_json = "1.0"
 signal-hook = "0.1"
 time = "0.1"
 
+[dependencies.cpuprofiler]
+# TODO(https://github.com/AtheMathmo/cpuprofiler/pull/10): Replace this with
+# 0.0.4 or an upstream branch reference once released.
+branch = "pkg-config"
+git = "https://github.com/jmmv/cpuprofiler.git"
+optional = true
+version = "0.0.3"
```
taralx: version doesn't do anything here. You should probably specify "rev" though.

src/main.rs
```diff
@@ -265,8 +267,14 @@ fn safe_main(program: &str, args: &[String]) -> Fallible<()> {
         return Err(UsageError { message: "invalid number of arguments".to_string() }.into());
     };
 
+    let profiler = match matches.opt_str("cpu_profile") {
```
taralx: I was going to suggest .map here, but you can actually do this with a conditional assignment:

```rust
let profiler;
if let Some(path) = matches.opt_str("cpu_profile") {
  profiler = sandboxfs::ScopedProfiler::start(&path).context("Failed to start CPU profile")?;
}
```

src/main.rs
```diff
@@ -265,8 +267,14 @@ fn safe_main(program: &str, args: &[String]) -> Fallible<()> {
...
+        Some(path) =>
+            Some(sandboxfs::ScopedProfiler::start(&path).context("Failed to start CPU profile")?),
+        None => None,
+    };
     sandboxfs::mount(mount_point, &options, &mappings, ttl, input, output)
         .context(format!("Failed to mount {}", mount_point.display()))?;
+    drop(profiler);
```
taralx: Why?

src/profiling.rs
```diff
@@ -0,0 +1,69 @@
...
+    #[cfg(not(feature = "profiling"))]
+    fn real_start<P: AsRef<Path>>(_path: P) -> Fallible<ScopedProfiler> {
+        Err(format_err!("Compile-time \"profiling\" feature not enabled"))
+    }
+
+    #[cfg(feature = "profiling")]
+    fn real_start<P: AsRef<Path>>(path: P) -> Fallible<ScopedProfiler> {
+        let path_str = match path.as_ref().to_str() {
+            Some(path_str) => path_str,
+            None => return Err(format_err!("Invalid path {}", path.as_ref().display())),
```
taralx: Don't need as_ref here.

integration/utils/config.go
```diff
@@ -17,10 +17,15 @@ package utils
...
 	"fmt"
 	"path/filepath"
+	"strings"
 )
 
 // Config represents the configuration for the integration tests as provided in the command line.
 type Config struct {
+	// features contains the set of features enabled during the build.  This represents a set
+	// and therefore the values in the map are meaningless.
+	features map[string]bool
```
taralx: Why unexported?

Cargo.toml
jmmv: Oh, of course. Replaced with `rev` and removed `branch`.

integration/utils/config.go
jmmv: Because I added the `HasFeature` method below to query this.

src/main.rs
jmmv: Done but...

I'm pretty confused about how this works as this makes it look as if `profiler` can be left uninitialized and then accessed. Is this only possible because `profiler` is never explicitly used?

src/main.rs
jmmv: Just to make it explicit that we are interested in profiling only the `mount()` loop. I can take it out though; don't feel strongly about it.

src/profiling.rs
jmmv: Looks like I do. If I remove it, the compiler complains that `P` doesn't have a `display()` method. And if I remove the `display()` call, then it complains about this not being printable with the default formatter.

src/main.rs
jmmv: Oh, but turns out I have to remove it in order for your comment above to work:

```
error[E0381]: use of possibly uninitialized variable: `profiler`                                                                                                                              
   --> src/main.rs:276:10                                                                                                                                                                     
    |                                                                                                                                                                                         
276 |     drop(profiler);                                                                                                                                                                     
    |          ^^^^^^^^ use of possibly uninitialized `profiler`                                                                                                                              
                                                                                                                                                                                              
error: aborting due to previous error                                                                                                                                                         
```

So removed.

jmmv:
ping! Sorry for the poke, but it'd be nice to be able to work on a release during this long weekend :)

integration/utils/config.go
taralx: Meh. I don't think you get anything out of using an accessor here, but up to you.

src/profiling.rs
taralx: Oh, right, P is AsRef. I feel like you want `let path = path.as_ref()`, but it's not critical.

src/main.rs
taralx: Correct.

src/profiling.rs
jmmv: Ah, that's a good idea. Done.

integration/utils/config.go
jmmv: Yeah, fair enough. Removed.

PR #68, opened 2019-01-29
Lint manpage

jmmv:
Spotted a few warnings in the manpage while preparing the release. We should have caught these earlier.


jmmv:
Yeah, that's how I noticed them too -- from your copy/paste in the release PR ;-)

PR #70, opened 2019-02-01
Make sandboxfs publishable as a crate

jmmv:
These changes are to avoid depending on unpublished code, which has bitten me at the last minute before being able to publish sandboxfs as a crate... and that have had me confused for a couple of days trying to find a solution (which was trivial in the end actually).

Further, these changes add linting of the crate to ensure we don't hit this again in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ go_import_path: github.com/bazelbuild/sandboxfs
 env:
   - DO=install
   - DO=lint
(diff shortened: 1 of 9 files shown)


jmmv:
Ping! This should be quite easy to review despite the multiple commits (`Cargo.toml` changes and a revert of a previous change), and is blocking 0.1 publication ;-)

PR #75, opened 2019-02-27
Use hashbrown for better performance

jmmv:
Profiling sandboxfs during large builds showed a non-negligible (but
not large) amount of time going into hashtable operations.  Given
that we don't care about "secure" hashing, we can afford to switch
to a faster implementation like hashbrown, so do it here.

This yields a ~1% performance boost during the Bazel build of a large
application using sandboxfs.  The following numbers are based on 10
consecutive builds for each case and are in seconds:

std:       mean 3429.56, median 3425.00, stddev 21.56
hashbrown: mean 3394.50, median 3378.50, stddev 36.31

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license = "Apache-2.0"
 name = "sandboxfs"
 readme = "README.md"
 repository = "https://github.com/bazelbuild/sandboxfs"
-version = "0.1.0"
+version = "0.1.1"
 
 [badges]
 travis-ci = { repository = "bazelbuild/sandboxfs", branch = "master" }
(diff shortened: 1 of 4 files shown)


djc:
(You're probably aware of this already, but in case anyone else is wondering about this, you actually switched from `std::collections::HashMap` with SipHash 1-3 to `hashbrown::HashMap` with FxHash -- so it's unclear whether the performance came from the change of implementation, the change of hash function, or (most likely) both.)

jmmv:
@djc Ah thanks. I noticed that I could make the standard HashMap use FxHash, but the numbers in the hashbrown page seemed convincing to switch both details at once :P I could measure again all these cases, but given that the performance differences are pretty minor and there is some noise in the metrics, it'd be hard to appreciate those.

djc:
Yeah, it makes sense to switch both at the same time in this case -- it's just that your commit message saying "Given that we don't care about "secure" hashing, we can afford to switch to a faster implementation like hashbrown, so do it here." does not entirely make sense.

jmmv:
Thanks @hlopko. Also updated the commit based on @djc's comment.

PR #78, opened 2019-03-29
Improve input/output handling

jmmv:


src/concurrent.rs
```diff
@@ -72,6 +72,16 @@ impl ShareableFile {
         }
     }
 
+    /// Constructs a new `ShareableFile` from an stdin handle and takes ownership of it.
+    pub fn from_stdin(stdin: io::Stdin) -> ShareableFile {
+        use std::os::unix::io::AsRawFd;
+        ShareableFile {
+            fd: stdin.as_raw_fd(),
```
taralx: Ehhhh, technically you're trying to steal it when it's not yours. Is it worth dup'ing it into a File so you don't have to propagate so much special case?

src/lib.rs
```diff
@@ -725,10 +728,17 @@ pub fn mount(mount_point: &Path, options: &[&str], mappings: &[Mapping], ttl: Ti
...
+        let mut input = match input {
+            Some(file) => concurrent::ShareableFile::from(file),
+            None => concurrent::ShareableFile::from_stdin(io::stdin()),
+        };
+        let reader = input.reader()?;
+
         let handler = thread::spawn(move || {
+            let writer: Box<io::Write> = match output {
+                Some(file) => Box::from(file),
+                None => Box::from(io::stdout()),
```
taralx: Please lock it. Otherwise you engage the lock every time you try to write to it.

src/concurrent.rs
jmmv: Uh, I didn't think about nor notice I could have created a `File` from the standard descriptors. Requires an `unsafe` block... but oh well, it seems much cleaner than what I did.

PTAL; this second commit is essentially redone now.

src/lib.rs
jmmv: Stale after rewrite per the comment above.

PR #80, opened 2019-04-13
Allow creating files as different users

jmmv:
When `--allow=other` is enabled, this ensures that files created in underlying paths have the right ownerships. The need for this has come up in the context of remote execution, where the environment there is... special for historical reasons.

I'm not sure if what I did here is the best way to achieve this, but it seems convoluted. I could think of two approaches: first, the one I went with (changing euid/egid at creation time), and, second, using `chown` right after creating the files. I went with the former approach under the assumption that the global lock is less costly than the doubled number of system calls... but I could be wrong.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,9 @@ failure = "~0.1.2"
 fuse = "0.3"
 getopts = "0.2"
 hashbrown = "0.1"
+lazy_static = "1.3.0"
 log = "0.4"
-nix = "0.12"
+nix = { git = "https://github.com/jmmv/nix.git", branch = "seteuid" }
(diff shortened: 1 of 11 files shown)


jmmv:
Huh... actually what I said is probably nonsensical. I was somehow thinking in terms of I/O syscalls all this time and got carried away by the coolness factor of that `run_as` function... but oh well, `sete[gu]id` are syscalls too, and we have to issue a bunch of them. Plus oh the complexity.

I'll try the `chown` approach separately to see how it looks like. Keeping this PR open just in case, but removing review request for now.

jmmv:
I think #81 is more sane than this, so closing.

PR #81, opened 2019-04-15
Allow creating files as different users

jmmv:
When --allow=other is enabled, this ensures that files created in underlying paths have the right ownerships. The need for this has come up in the context of remote execution, where the environment there is... special for historical reasons.

Obsoletes #80.

### DIFF
--- a/integration/read_write_test.go
+++ b/integration/read_write_test.go
@@ -49,6 +49,63 @@ func openAndDelete(path string, mode int) (int, error) {
 	return fd, nil
 }
 
+// createAsDifferentUser implements a generic test that creates entries within the mount point
+// as a different user than the one running sandboxfs and checks that the created entries in the
+// underlying tree has the right owner:group settings.
+//
+// The createAsUser lambda is a function responsible for creating the desired type of entry at the
+// given path with the given credentials.
+func createAsDifferentUserTest(t *testing.T, createAsUser func(string, *utils.UnixUser) error) {
+	root := utils.RequireRoot(t, "Requires root privileges")
+
+	user := utils.GetConfig().UnprivilegedUser
+	if user == nil {
+		t.Skipf("unprivileged user not set; must contain the name of an unprivileged user with FUSE access")
+	}
+	t.Logf("Using primary unprivileged user: %v", user)
+
+	state := utils.MountSetupWithUser(t, root, "--mapping=rw:/:%ROOT%", "--allow=other")
+	defer state.TearDown(t)
+
+	utils.MustMkdirAll(t, state.RootPath("dir"), 0777)
+	if err := os.Chown(state.RootPath("dir"), user.UID, user.GID); err != nil {
+		t.Fatalf("Cannot create unprivileged work directory: %v", err)
+	}
+
+	wantFiles := []struct {
+		path string
+		user *utils.UnixUser
+	}{
+		{path: "dir/unprivileged", user: user},
+		{path: "privileged", user: root},
+	}
+	for _, wantFile := range wantFiles {
+		if err := createAsUser(state.MountPath(wantFile.path), wantFile.user); err != nil {
+			t.Fatalf("Cannot create %s as %v: %v", wantFile.path, wantFile.user, err)
+		}
+	}
+
+	// Now that all files were created, make sure they have the right ownerships.
+	for _, wantFile := range wantFiles {
+		// Check that the underlying file has the right ownership, and also check that the
+		// node we recorded in sandboxfs' memory agrees.
+		for _, path := range []string{state.MountPath(wantFile.path), state.RootPath(wantFile.path)} {
+			fileInfo, err := os.Lstat(path)
+			if err != nil {
+				t.Fatalf("Cannot stat %s: %v", path, err)
+			}
+			stat := fileInfo.Sys().(*syscall.Stat_t)
+			if int(stat.Uid) != wantFile.user.UID || int(stat.Gid) != wantFile.user.GID {
+				t.Errorf("%s has wrong ownership; got %v:%v, want %v:%v", path, stat.Uid, stat.Gid, wantFile.user.UID, wantFile.user.GID)
+			}
+		}
+	}
+}
+
+func TestReadWrite_MkdirAsDifferentUser(t *testing.T) {
+	createAsDifferentUserTest(t, utils.MkdirAsUser)
+}
+
 func TestReadWrite_CreateFile(t *testing.T) {
 	state := utils.MountSetup(t, "--mapping=rw:/:%ROOT%")
 	defer state.TearDown(t)
@@ -63,6 +120,10 @@ func TestReadWrite_CreateFile(t *testing.T) {
 	if err := utils.FileEquals(state.MountPath("subdir/file"), "new content"); err != nil {
 		t.Error(err)
 	}
+}
+
+func TestReadWrite_CreateFileAsDifferentUser(t *testing.T) {
+	createAsDifferentUserTest(t, utils.CreateFileAsUser)
 }
 
 func TestReadWrite_Remove(t *testing.T) {
@@ -326,8 +387,6 @@ func equivalentStats(stat1 os.FileInfo, stat2 os.FileInfo) error {
 // Tests calling this function should only start a sandboxfs instance with the desired configuration
 // and then immediately call this function.
 func doRenameTest(t *testing.T, oldOuterPath, newOuterPath, oldInnerPath, newInnerPath string) {
-	t.Helper()
-
 	utils.MustMkdirAll(t, filepath.Dir(oldOuterPath), 0755)
 	utils.MustMkdirAll(t, filepath.Dir(newOuterPath), 0755)
 	utils.MustMkdirAll(t, filepath.Dir(oldInnerPath), 0755)
@@ -493,6 +552,15 @@ func TestReadWrite_MoveRace(t *testing.T) {
 	}
 }
 
+func TestReadWrite_MoveAsDifferentUser(t *testing.T) {
+	createAsDifferentUserTest(t, func(path string, user *utils.UnixUser) error {
+		if err := utils.CreateFileAsUser(path+".old", user); err != nil {
+			return err
+		}
+		return utils.MoveAsUser(path+".old", path, user)
+	})
+}
+
 func TestReadWrite_Mknod(t *testing.T) {
 	utils.RequireRoot(t, "Requires root privileges to create arbitrary nodes")
 
@@ -595,6 +663,9 @@ func TestReadWrite_Mknod(t *testing.T) {
 			}
 		})
 	}
+}
+func TestReadWrite_MknodAsDifferentUser(t *testing.T) {
+	createAsDifferentUserTest(t, utils.MkfifoAsUser)
 }
 
 func TestReadWrite_Chmod(t *testing.T) {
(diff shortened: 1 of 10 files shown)


integration/read_write_test.go
```diff
@@ -49,6 +49,66 @@ func openAndDelete(path string, mode int) (int, error) {
...
+
+	state := utils.MountSetupWithUser(t, root, "--mapping=rw:/:%ROOT%", "--allow=other")
+	defer state.TearDown(t)
+
+	utils.MustMkdirAll(t, state.RootPath("dir"), 0777)
+	if err := os.Chown(state.RootPath("dir"), user.UID, user.GID); err != nil {
+		t.Fatalf("Cannot create unprivileged work directory: %v", err)
+	}
+
+	type WantFile struct {
```
taralx:
```go
testcases := []struct{
  ...
}{
  {...},
  ...
}

for _, tc := range testcases {
```

integration/read_write_test.go
```diff
@@ -49,6 +49,66 @@ func openAndDelete(path string, mode int) (int, error) {
...
 }
 
+// createAsDifferentUser implements a generic test that creates entries within the mount point
+// as a different user than the one running sandboxfs and checks that the created entries in the
+// underlying tree has the right owner:group settings.
+//
+// The createAsUser lambda is a function responsible for creating the desired type of entry at the
+// given path with the given credentials.
+func createAsDifferentUserTest(t *testing.T, createAsUser func(string, *utils.UnixUser) error) {
+	t.Helper()
```
taralx: This is far too complex to count as a helper.

src/lib.rs
```diff
@@ -456,16 +457,53 @@ impl SandboxFS {
...
+/// The file is created via the `create` lambda, which can create any type of file it wishes.  The
+/// `delete` lambda should match this creation and allow the deletion of the file, and this is used
+/// as a cleanup function when the ownership cannot be successfully changed.
+fn create_as<T, E, P, Create, Delete>(
+    path: &P, uid: unistd::Uid, gid: unistd::Gid, create: Create, delete: Delete) -> Result<T, E>
+    where P: AsRef<Path>, Create: Fn(&P) -> Result<T, E>, Delete: Fn(&P) -> Result<(), E>,
+          E: From<Errno> {
+    let mut res = create(path);
+    if res.is_ok() {
+        if let Err(e) = unistd::fchownat(None, path.as_ref(), Some(uid), Some(gid),
```
taralx: Not worried about races?

src/lib.rs
```diff
@@ -456,16 +457,53 @@ impl SandboxFS {
...
+/// Creates a file `path` with the given `uid`/`gid` pair.
+///
+/// The file is created via the `create` lambda, which can create any type of file it wishes.  The
+/// `delete` lambda should match this creation and allow the deletion of the file, and this is used
+/// as a cleanup function when the ownership cannot be successfully changed.
+fn create_as<T, E, P, Create, Delete>(
+    path: &P, uid: unistd::Uid, gid: unistd::Gid, create: Create, delete: Delete) -> Result<T, E>
+    where P: AsRef<Path>, Create: Fn(&P) -> Result<T, E>, Delete: Fn(&P) -> Result<(), E>,
+          E: From<Errno> {
+    let mut res = create(path);
```
taralx: Use ? syntax.

src/lib.rs
```diff
@@ -456,16 +457,53 @@ impl SandboxFS {
...
+    if res.is_ok() {
+        if let Err(e) = unistd::fchownat(None, path.as_ref(), Some(uid), Some(gid),
+                                         unistd::FchownatFlags::NoFollowSymlink) {
+            res = Err(E::from(e.as_errno().expect("fchownat should have failed with an errno")));
+            if let Err(delete_err) = delete(path) {
+                warn!("Cannot delete just-created file {} after failing to change ownership",
+                      path.as_ref().display());
+                if res.is_ok() {
```
taralx: This is always false.

src/lib.rs
```diff
@@ -456,16 +457,53 @@ impl SandboxFS {
...
+/// as a cleanup function when the ownership cannot be successfully changed.
+fn create_as<T, E, P, Create, Delete>(
+    path: &P, uid: unistd::Uid, gid: unistd::Gid, create: Create, delete: Delete) -> Result<T, E>
+    where P: AsRef<Path>, Create: Fn(&P) -> Result<T, E>, Delete: Fn(&P) -> Result<(), E>,
+          E: From<Errno> {
+    let mut res = create(path);
+    if res.is_ok() {
+        if let Err(e) = unistd::fchownat(None, path.as_ref(), Some(uid), Some(gid),
+                                         unistd::FchownatFlags::NoFollowSymlink) {
+            res = Err(E::from(e.as_errno().expect("fchownat should have failed with an errno")));
```
taralx: Prefer to handle the weird instead of panicking.

jmmv: Dubious about this. I don't think any of the other non-errno errors in nix are possible here, so getting them would mean that our logic is wrong -- and in those cases I favor crashing than returning an error.

But I'm not fully sure, so done. Doing this... made me change the structure of this function significantly. Not sure if it's now easier to read or worse than before.

taralx:
Hrm, it's better if you don't force push during review, because now I have no way of only seeing changes since last review. :(

src/lib.rs
```diff
@@ -890,4 +935,64 @@ mod tests {
...
+
+        let config = testutils::Config::get();
+        match config.unprivileged_user {
+            Some(user) => {
+                let uid = unistd::Uid::from_raw(user.uid());
+                let gid = unistd::Gid::from_raw(user.primary_group_id());
+                do_create_as_ok_test(uid, gid);
+            },
+            None => {
+                panic!("UNPRIVILEGED_USER must be set when running as root for this test to run");
```
taralx: Skip instead?

src/lib.rs
```diff
@@ -456,16 +457,59 @@ impl SandboxFS {
     }
 }
 
+/// Creates a file `path` with the given `uid`/`gid` pair.
+///
+/// The file is created via the `create` lambda, which can create any type of file it wishes.  The
+/// `delete` lambda should match this creation and allow the deletion of the file, and this is used
+/// as a cleanup function when the ownership cannot be successfully changed.
+fn create_as<T, E, P, Create, Delete>(
+    path: &P, uid: unistd::Uid, gid: unistd::Gid, create: Create, delete: Delete) -> Result<T, E>
```
taralx: I think this would be more readable if you used impl trait:

```rust
fn create_as<P: AsRef<Path>, T, E: From<Errno>>(path: &P, uid: unistd::Uid, gid: unistd::Gid, create: impl Fn(&P) -> Result<T, E>, delete: impl Fn(&P) -> Result<(), E>) -> Result<T, E>
```

(or use the where clause if you prefer it)

src/lib.rs
```diff
@@ -707,7 +751,7 @@ impl reconfig::ReconfigurableFS for ReconfigurableSandboxFS {
 /// Mounts a new sandboxfs instance on the given `mount_point` and maps all `mappings` within it.
 pub fn mount(mount_point: &Path, options: &[&str], mappings: &[Mapping], ttl: Timespec,
     input: fs::File, output: fs::File) -> Fallible<()> {
-    let mut os_options = options.iter().map(|o| o.as_ref()).collect::<Vec<&OsStr>>();
+    let mut os_options = options.iter().map(std::convert::AsRef::as_ref).collect::<Vec<&OsStr>>();
```
taralx: Is AsRef not imported here? Not sure this is better than the lambda, but up to you.

src/lib.rs
jmmv: It helps. Done.

src/lib.rs
```diff
@@ -456,16 +457,59 @@ impl SandboxFS {
...
+fn create_as<T, E, P, Create, Delete>(
+    path: &P, uid: unistd::Uid, gid: unistd::Gid, create: Create, delete: Delete) -> Result<T, E>
+    where P: AsRef<Path>, Create: Fn(&P) -> Result<T, E>, Delete: Fn(&P) -> Result<(), E>,
+          E: From<Errno> {
+    create(path).or_else(|create_err| Err({
+        match unistd::fchownat(None, path.as_ref(), Some(uid), Some(gid),
+                               unistd::FchownatFlags::NoFollowSymlink) {
+            Ok(()) => create_err,
+            Err(nix::Error::Sys(chown_errno)) => {
+                match delete(path) {
```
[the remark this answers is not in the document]
jmmv: Yeah, I don't know what I was thinking when writing this. Redone. Hopefully it's clear this time around.

src/lib.rs
```diff
@@ -456,16 +457,59 @@ impl SandboxFS {
...
+                    Ok(_) => E::from(chown_errno),
+                    Err(delete_err) => {
+                        warn!("Cannot delete created file {} after failing to change ownership",
+                              path.as_ref().display());
+                        delete_err
```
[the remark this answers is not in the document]
jmmv: Agreed. Oversight caused by previous complexity.

src/lib.rs
jmmv: Imported indeed. Done.

src/lib.rs
jmmv: Yes, I'd like to... but how though? I couldn't find a way with the default test runner.

taralx: Hrm, hrm, maybe just print something and return success? I dunno, this is probably ok.

src/lib.rs
```diff
@@ -462,31 +462,33 @@ impl SandboxFS {
...
-            Err(unknown_chown_error) => {
-                warn!("fchownat({}) failed with unexpected non-errno error: {:?}",
-                      path.as_ref().display(), unknown_chown_error);
-                E::from(Errno::EIO)
-            },
-        }
-    }))
+fn create_as<T, E: From<Errno> + fmt::Display, P: AsRef<Path>>(
+    path: &P, uid: unistd::Uid, gid: unistd::Gid,
+    create: impl Fn(&P) -> Result<T, E>, delete: impl Fn(&P) -> Result<(), E>) -> Result<T, E> {
```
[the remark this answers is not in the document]
jmmv: Probably, but I haven't been using rustfmt... Modified by hand.

src/lib.rs
```diff
@@ -462,31 +462,33 @@ impl SandboxFS {
...
+
+    let result = create(path)?;
+
+    match unistd::fchownat(
```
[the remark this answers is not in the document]
jmmv: Yeah, seems nicer this way.

src/lib.rs
```diff
@@ -462,31 +462,33 @@ impl SandboxFS {
...
+        None, path.as_ref(), Some(uid), Some(gid), unistd::FchownatFlags::NoFollowSymlink) {
+        Ok(()) => Ok(result),
+        Err(e) => {
+            let chown_errno = match e {
```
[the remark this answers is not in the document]
jmmv: With the conversion to `map_err` and the use of the ? operator, neither `from` nor `into` are explicitly required.

src/lib.rs
jmmv: I see. The reason I went with the panic is to make this very obvious and to prevent my CI scripts from missing this case by mistake. I don't expect anyone running these tests as root by hand. Further, `cargo test` swallows the output of successful tests by default, so the printed message wouldn't be visible (tried that before) :-/

PR #82, opened 2019-06-19
Fix build on 32-bit systems.

### LINKED ISSUE #71: Fails to build on FreeBSD/i386
```
error[E0308]: mismatched types
  --> src/nodes/conv.rs:72:19
   |
72 |     Timespec::new(val.tv_sec() as sys::time::time_t, usec)
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected i64, found i32
help: you can cast an `i32` to `i64`, which will sign-extend the source value
   |
72 |     Timespec::new((val.tv_sec() as sys::time::time_t).into(), usec)
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error
```

abenson:
I'm not well versed in rust, but seems correct to me and fixes the build issues on 32-bit.

Should Close #71.

### DIFF
--- a/src/nodes/conv.rs
+++ b/src/nodes/conv.rs
@@ -69,7 +69,7 @@ pub fn timeval_to_timespec(val: sys::time::TimeVal) -> Timespec {
     } else {
         val.tv_usec() as i32
     };
-    Timespec::new(val.tv_sec() as sys::time::time_t, usec)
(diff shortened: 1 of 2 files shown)


abenson:
I signed it!

PR #115, opened 2020-10-01
Check in Cargo.lock?

jeremyschlatter:
I was [packaging up sandboxfs](https://github.com/NixOS/nixpkgs/pull/99353) to make it available through the nix package manager, and stumbled on the lack of a checked-in Cargo.lock.

Checking in Cargo.lock seems to be the [recommended best practice](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) for binaries like sandboxfs, and would also make my life easier here, so I figured I would see if you were up for it.


classabbyamp:
+1, we also enforce lockfiles on void (e.g. `cargo --locked`)

jeremyschlatter:
Closing because this repo is unmaintained.

